PR #25, opened 2015-09-09
Revised Swift 2.0 support

mrackwitz:
Replaces the non-generic home-grown Result type with the in this case slightly more expressive, but less verbose syntax sugar given by Swift's error handling.


kylef:
This won't merge since it conflicts with the already existing Swift 2.0 support. Can you rebase it?

mrackwitz:
Wow :disappointed:

PR #29, opened 2015-09-30
bugfix for loading Template from URL

chunkerchunker:
NSURL.absoluteString includes "file://", which Path() doesn't expect.


kylef:
Ah, that's a good point. Thanks for the pull request.

PR #33, opened 2015-10-18
Adding Guards 👮

AliSoftware:
This PR goes on top of #32 and convert the code to modern Swift by using `guard` to make it waaayy more readable :wink: 

I kept #32 separated from this PR as they are not directly related and you might want to review them separately, but this PR still requires #32 to be merged first — as the added `guard` statements also apply to changes made by #32

### DIFF
--- a/Stencil/Node.swift
+++ b/Stencil/Node.swift
@@ -118,27 +118,27 @@ public class ForNode : NodeType {
   public class func parse(parser:TokenParser, token:Token) throws -> NodeType {
     let components = token.components()
 
-    if components.count == 4 && components[2] == "in" {
-      let loopVariable = components[1]
-      let variable = components[3]
-
-      var emptyNodes = [NodeType]()
-
-      let forNodes = try parser.parse(until(["endfor", "empty"]))
-
-      if let token = parser.nextToken() {
-        if token.contents == "empty" {
-          emptyNodes = try parser.parse(until(["endfor"]))
-          parser.nextToken()
-        }
-      } else {
-        throw TemplateSyntaxError("`endfor` was not found.")
-      }
-
-      return ForNode(variable: variable, loopVariable: loopVariable, nodes: forNodes, emptyNodes:emptyNodes)
+    guard components.count == 4 && components[2] == "in" else {
+      throw TemplateSyntaxError("'for' statements should use the following 'for x in y' `\(token.contents)`.")
     }
-
-    throw TemplateSyntaxError("'for' statements should use the following 'for x in y' `\(token.contents)`.")
+    
+    let loopVariable = components[1]
+    let variable = components[3]
+    
+    var emptyNodes = [NodeType]()
+    
+    let forNodes = try parser.parse(until(["endfor", "empty"]))
+    
+    guard let token = parser.nextToken() else {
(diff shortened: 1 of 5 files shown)


Stencil/Node.swift
```diff
@@ -73,10 +73,13 @@ public class VariableNode : NodeType {
 public class NowNode : NodeType {
   public let format:Variable
 
-  public class func parse(parser:TokenParser, token:Token) -> NodeType {
+  public class func parse(parser:TokenParser, token:Token) throws -> NodeType {
     var format:Variable?
 
     let components = token.components()
+    guard components.count <= 2 else {
+      throw TemplateSyntaxError("'now' tags should can only have one argument: the format string `\(token.contents)`.")
```
kylef: Please switch `can` to `may` (RFC 2119)

kylef:
This PR still includes the changes for #32, can you remove these from this PR please.

Stencil/Node.swift
```diff
@@ -118,33 +117,37 @@ public class ForNode : NodeType {
...
 
-    throw TemplateSyntaxError("'for' statements should use the following 'for x in y' `\(token.contents)`.")
+    return ForNode(variable: variable, loopVariable: loopVariable, nodes: forNodes, emptyNodes: emptyNodes)
   }
 
   public init(variable:String, loopVariable:String, nodes:[NodeType], emptyNodes:[NodeType]) {
     self.variable = Variable(variable)
     self.loopVariable = loopVariable
     self.nodes = nodes
+    // TODO: Handle emptyNodes
```
kylef: Good spot

Stencil/TemplateLoader/Inheritence.swift
```diff
@@ -58,22 +58,20 @@ class ExtendsNode : NodeType {
...
-
-      let paths:String = loader.paths.map { path in
-        return path.description
-        }.joinWithSeparator(", ")
+    guard let loader = context["loader"] as? TemplateLoader else {
+      throw TemplateSyntaxError("Template loader not in context")
+    }
+    
+    guard let template = loader.loadTemplate(templateName) else {
+      let paths:String = loader.paths.map { path in return path.description }.joinWithSeparator(", ")
```
kylef: `return` here is unnecessary since single line closures have implicit return

AliSoftware:
> This PR still includes the changes for #32, can you remove these from this PR please.

Not really actually, as they would conflict if separated: there are `guard` statements added there that are on the same code regions that the ones where I added `parser.eatNextCRLF()` calls. Splitting those PRs would make the merge conflict much harder ^^

Stencil/Node.swift
AliSoftware: I wonder why the compiler didn't warn about unused variable.parameter btw, I'd expect it to suggest you a _Fix-It_ to replace `emptyNodes` with `_` in such cases.

AliSoftware:
You can isolate [the diff between #32 and this PR here](https://github.com/AliSoftware/Stencil/compare/fix/22-extra_lines...AliSoftware:guards) if you want to review it more easily

AliSoftware:
Ok, this is now entirely rewritten without the commits from #32 and ready to merge!

kylef:
Thanks for the pull request!

PR #36, opened 2015-10-26
Fix IfNode when using Array of arbitrary types

AliSoftware:
which made the cast to [AnyObject] fail.

See also AliSoftware/SwiftGen#41 that was affected by that issue.

PS: No `CHANGELOG.md`?


kylef:
Thanks for digging into this and fixing the underlying problem.

@AliSoftware The change log can be found at https://github.com/kylef/Stencil/releases. This particular change won't be mentioned in the changelog because this wasn't broken in the last release, only in master.

PR #38, opened 2015-11-03
Add context property to template

Danappelxx:
It's a very small change, but it allows context to be stored in the template, meaning that it doesn't always have to be given upon rendering. It's not terribly useful but I need the change for a project I'm working on, maybe others do too :)

### DIFF
--- a/Stencil/Template.swift
+++ b/Stencil/Template.swift
@@ -4,6 +4,7 @@ import PathKit
 /// A class representing a template
 public class Template {
   public let parser:TokenParser
+  public var context:Context? = nil
   private var nodes:[NodeType]? = nil
 
   /// Create a template with the given name inside the given bundle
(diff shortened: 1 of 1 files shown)


Danappelxx:
Ping?

kylef:
I'm not sure I understand the use case, are you wishing to call render multiple times with the same context?

kylef:
Currently a `Template` is immutable and I'd prefer to keep it that way, unless there is a good reason for making templates mutable.

Danappelxx:
Well in my case I'm trying to make it conform to a general protocol

```swift
protocol Template {
    var context: {get}
    func render() throws -> String
}
```
And this is impossible to do without the context property.

This would make templates mutable though, which isn't a great thing.

Danappelxx:
After giving it some thought I think this is a bad idea for the reasons you gave. I think I can achieve my goal with partially applied arguments, anyway.

PR #40, opened 2015-11-20
Cast ForNode values to Array<Any>

njdehoog:
If `Context` contains an array of type `Array<Any>` the cast to NSArray in the `render` function of `ForNode` fails, causing the `ForNode` to render nothing. This PR addresses that issue by mapping `values` to an array of type `Array<Any>`.

The following spec illustrates the scenario that would fail before this change:

```swift
  $0.it("renders a context variable of type Array<Any>") {
    let any_context = Context(dictionary: [
        "items": ([1, 2, 3] as [Any])
      ])

    let nodes: [NodeType] = [VariableNode(variable: "item")]
    let node = ForNode(variable: "items", loopVariable: "item", nodes: nodes, emptyNodes: [])
    try expect(try node.render(any_context)) == "123" 
  }
```

### DIFF
--- a/Stencil/Node.swift
+++ b/Stencil/Node.swift
@@ -157,7 +157,7 @@ public class ForNode : NodeType {
   public func render(context: Context) throws -> String {
     let values = try variable.resolve(context)
 
-    if let values = values as? NSArray where values.count > 0 {
+    if let values = values as? [Any] where values.count > 0 {
       return try values.map { item in
         try context.push([loopVariable: item]) {
           try renderNodes(nodes, context)
@@ -232,10 +232,10 @@ public class IfNode : NodeType {
     let result = try variable.resolve(context)
     var truthy = false
 
-    if let result = result as? NSArray {
-      if result.count > 0 {
-        truthy = true
-      }
+    if let result = result as? [Any] {
+      truthy = !result.isEmpty
+    } else if let result = result as? [String:Any] {
+      truthy = !result.isEmpty
     } else if result != nil {
       truthy = true
     }
--- a/Stencil/Variable.swift
+++ b/Stencil/Variable.swift
@@ -58,11 +58,9 @@ public struct Variable : Equatable, Resolvable {
     for bit in lookup() {
       if let context = current as? Context {
         current = context[bit]
-      } else if let dictionary = current as? [String: Any] {
+      } else if let dictionary = resolveDictionary(current) {
         current = dictionary[bit]
-      } else if let dictionary = current as? [String: AnyObject] {
-        current = dictionary[bit]
-      } else if let array = current as? [Any] {
+      } else if let array = resolveArray(current) {
         if let index = Int(bit) {
           current = array[index]
         } else if bit == "first" {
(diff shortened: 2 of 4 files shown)


AliSoftware:
Careful I had an issue before at a time when Stencil didn't use `NSArray` but `[Any]` or `[AnyObject]` and that cas did fail in other circumstances, that's actually one reason why it has been switched to `NSArray`

See #36 (that was for the `IfNode` but probably the same or related issue)

njdehoog:
Thanks for pointing that out @AliSoftware. Could you specify in what case exactly this would fail?

The problem with trying to cast to `NSArray` is that this does not always work for values of type `Array`, since `NSArray` can only contain reference types.

kylef:
What @AliSoftware mentions is already tested against so you didn't break that case.

See  https://github.com/kylef/Stencil/commit/ba65ab5fbeb78fb2e8e7e4d34703c7f390738e8c#diff-477705f51da7594200b181ade10bba3bR86

njdehoog:
Good point. Probably this should be fixed for for the `IfNode` case as well though. Investigating now.

AliSoftware:
yeah I see that you did not simply try to cast to `[Any]` but kept the cast to `NSArray` too so as long as you do both this should be OK.

The specific type where it failed is detailed in my comment on swiftgen linked above: my context entry was of type `[Swift.Dictionary<Swift.String, Swift.AnyObject>]` which should be castable to `Array<Any>` in theory, but in practice wasn't (Swift bug?)

At that time I added the failing specs [here](https://github.com/kylef/Stencil/pull/36/files). Those two additional specs did fail before my fix but strangely also did fail when I tried to fix it by casting to `[Any]` (instead of `[AnyObject]`) so that's why I had to cast to `NSArray` to make it work (even if `Swift.Dictionary` is a value type, but thanks to the `import Foundation` it's auto-bridged to `NSDictionary` in that case and cast to `NSArray` succeeds).

You might want to duplicate my two specs in the `ForNodeSpec` to ensure that doesn't break. Maybe we should also test multiple casts like you did in this PR (instead of choosing either `NSArray` or `Array<Any>`, test them both like you did) but for all the nodes, not only `ForNode`.

AliSoftware:
> What @AliSoftware mentions is already tested against so you didn't break that case.

@kylef not entirely true, those specs were testing the `IfNode`, which casts to `NSArray`, but the modification here is in the `ForNode`. So we should duplicate my specs in `IfNodeSpec` to have them in `ForNodeSpec` too, and we should also apply the same modifications done here in `ForNode` for `IfNode` (test both a conversion to `NSArray` and to `Array<Any>` instead of only `NSArray`)

njdehoog:
So the same issue occurred in `IfNode`. See this spec:

```swift
$0.it("renders false when Array<Any> variable is empty") {
  let arrayContext = Context(dictionary: ["items": ([] as [Any])])
  let node = IfNode(variable: "items", trueNodes: [TextNode(text: "true")], falseNodes: [TextNode(text: "false")])
  try expect(try node.render(arrayContext)) == "false"
}
```

I've added the same `flatMap` function in `IfNode`, but since we're now duplicating the same code it should probably be generalized. Any suggestions on a clean way to do this @kylef?

kylef:
I think it would be best to make Variable normalise these values when resolving variables, so each node doesn't need to care or know about these details. I've made these changes in 22919dc (branch: resolve), would you be able to test this behaviour @njdehoog? You could rebase your test-cases on-top of these changes to test.

njdehoog:
Thanks @kylef. That looks good. I've rebased on top of `resolve`. The tests still pass.

I've also changed the syntax slightly in the resolve functions. I feel like a `switch` statement more clearly expresses the intent in this case (no pun intended).

Are we good to merge to master?

kylef:
Great, thanks @njdehoog.

PR #41, opened 2015-11-21
Replace NSRegularExpression with string scanning

kylef:
This removes the hard-dependency on Foundation, the only dependencies for Foundation remaining are the `NowNode` and support for key-value-coding, and NSBundle loading, which can be wrapped around conditional compilation once Swift is available for other platforms.

### DIFF
--- a/Stencil/Lexer.swift
+++ b/Stencil/Lexer.swift
@@ -1,16 +1,13 @@
-import Foundation
-
 public struct Lexer {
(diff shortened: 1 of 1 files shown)


kattrali:
This looks good. I was looking for some edge cases with nested tokens or something, but none of those make any real world sense. It makes sense to eventually split out `Scanner` though.

PR #43, opened 2015-11-23
Allow whitespace in filter expression

### LINKED ISSUE #42: Template filters should allow whitespace
Filters currently only work when there is no whitespace surrounding the pipe character. It would be a lot nicer if whitespace was stripped from the variable names surrounding the pipe character.

This doesn't work at the moment:
```swift
{{ post.title | uppercase }}
```

njdehoog:
Fixes #42. Spec included.


kylef:
Thanks for this pull request.

PR #44, opened 2015-11-24
Allow filter arguments

njdehoog:
I would like to propose an addition to custom filters functionality, which allows arguments to be passed to filter functions, just like in [Liquid](https://docs.shopify.com/themes/liquid-documentation/basics). This PR represents a work in progress implementation of that functionality.

Example syntax:

```
{{ post.date | date: "EEE, dd MMM yyyy" }}
```
### Syntax

The syntax I want to propose is as follows:
{{ `variable` | `filter_name`: `arg1`, `arg2` }}

While taking the following rules into account:
- Whitespace is stripped from filter name and/or arguments, unless argument is in quotes
- If argument is in quotes it is treated as a string literal
- If argument is not in quotes it is treated as a variable (which should be resolved before being passed to the function), unless it is a number
### Improvements
- Current implementation treats all arguments as string literals
- Current implementation requires all filter functions to accept arguments. It would be great if that could be optional, and it would be even better if arguments could be passed as actual arguments to the function instead of as an array.

Any comments/suggestions would be appreciated.

### DIFF
--- a/Stencil/Filters.swift
+++ b/Stencil/Filters.swift
@@ -8,7 +8,7 @@ func toString(value: Any?) -> String? {
   return nil
 }
 
-func capitalise(value: Any?, arguments: [Any?]) -> Any? {
+func capitalise(value: Any?) -> Any? {
   if let value = toString(value) {
     return value.capitalizedString
   }
@@ -16,7 +16,7 @@ func capitalise(value: Any?, arguments: [Any?]) -> Any? {
   return value
 }
 
-func uppercase(value: Any?, arguments: [Any?]) -> Any? {
+func uppercase(value: Any?) -> Any? {
   if let value = toString(value) {
     return value.uppercaseString
   }
@@ -24,7 +24,7 @@ func uppercase(value: Any?, arguments: [Any?]) -> Any? {
   return value
 }
 
-func lowercase(value: Any?, arguments: [Any?]) -> Any? {
+func lowercase(value: Any?) -> Any? {
   if let value = toString(value) {
     return value.lowercaseString
   }
--- a/Stencil/Filters.swift
+++ b/Stencil/Filters.swift
@@ -8,7 +8,7 @@ func toString(value: Any?) -> String? {
   return nil
 }
 
-func capitalise(value: Any?, args: [String]?) -> Any? {
+func capitalise(value: Any?, args: [Any?]) -> Any? {
   if let value = toString(value) {
     return value.capitalizedString
   }
@@ -16,7 +16,7 @@ func capitalise(value: Any?, args: [String]?) -> Any? {
   return value
 }
 
-func uppercase(value: Any?, args: [String]?) -> Any? {
+func uppercase(value: Any?, args: [Any?]) -> Any? {
   if let value = toString(value) {
     return value.uppercaseString
   }
@@ -24,7 +24,7 @@ func uppercase(value: Any?, args: [String]?) -> Any? {
   return value
 }
 
-func lowercase(value: Any?, args: [String]?) -> Any? {
+func lowercase(value: Any?, args: [Any?]) -> Any? {
   if let value = toString(value) {
     return value.lowercaseString
   }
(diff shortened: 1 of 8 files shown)


njdehoog:
The last commit makes PR #43 obsolete.

kylef:
Hi @njdehoog, I think we should instead be using `Variable`s throughout, this will be more flexible and will reduce a large portion of the duplicated code related to `String.smartSplit`. A filter should then take Any as it can't make the assumption the value will be a string.

Example:

```html+django
{{ name|default:"unknown user" }}
{{ avatar|default:defaultAvatar }}
```

```swift
func default(value: Any?, arguments: [Any]) throws -> Any? {
  guard arguments.count == 1 else {
    throw /* must be 1 argument */
  }

  if let value = value {
    return value
  }

  return arguments.first
}
```

Where `defaultAvatar` is a variable and `"unknown user"` is a literal value. [Variable](https://github.com/kylef/Stencil/blob/master/Stencil/Variable.swift#L53-L56) handles string literals itself.

Would you be able to adapt this to use Variable instead?

Stencil/Filters.swift
```diff
@@ -8,7 +8,7 @@ func toString(value: Any?) -> String? {
   return nil
 }
 
-func capitalise(value: Any?) -> Any? {
+func capitalise(value: Any?, args: [String]?) -> Any? {
```
kylef: Since filters with no arguments will be more common, I think we should have a `registerFilter` function inside the `Namespace` which will accept filters without arguments so this change would be unnecessary. It should wrap the given filter in a closure which accepts the arguments and throws if arguments have been presented.

Stencil/Parser.swift
```diff
@@ -10,7 +10,7 @@ public func until(tags:[String])(parser:TokenParser, token:Token) -> Bool {
   return false
 }
 
-public typealias Filter = Any? throws -> Any?
+public typealias Filter = (value: Any?, args: [String]?) throws -> Any?
```
kylef: Please don't use be an abbreviated variant of `arguments`.

njdehoog:
@kylef thanks for your comments. I changed the code to use `Variable` for filter arguments, but then I ran into the problem that `Variable` doesn't (need to) know about integer literals. So in order to support integer literals as filter arguments, I created a new `FilterArgument` type. I'm not sure if this is the best approach, so please let me know if you have a better idea.

njdehoog:
I created a `Filter` enum type to support both filters with and without arguments. Unfortunately this slightly changes the syntax for registering a filter:

```swift
let filter = Filter() { value in
  if let value = value as? String {
    return "\(value) \(value)"
  }

  return nil
}
namespace.registerFilter("repeat", filter: filter)
```

The `Filter` type can be instantiated with a closure that accepts a value, or a closure that accepts both value and arguments. I tried to maintain the current syntax by overloading `registerFilter` for the two different function types and have that method create a `Filter` instance, but in that case the Swift compiler is unable to infer the type of closure and throws this error:

```
Ambiguous use of 'registerFilter(_:filter:)'
```

I'm not sure why overloading works for the initializers on `Filter` but not for the `registerFilter` method.

What do you think @kylef? Is this an acceptable solution?

njdehoog:
@kylef I rebased on master and I'd say this is ready to be merged. Any more feedback?

kylef:
Sorry for the huge delay in getting back to this one, for some reason I missed the notification for your message. Greatly sorry for not checking back.

I'll rebase this myself on master over the weekend and get this in.

Thanks for the PR.

njdehoog:
Sounds good @kylef. Looking forward to seeing this merged in. Stencil has become an important part of the app I'm working on, and I'd love to keep contributing the things that I'm adding to it back to the main repository.

kylef:
Sorry for taking so long on this one. I've added this feature in 60b378d482011b41bc630f85a8b3f931a55e434d where I've made some changes (namely removing the filter enum from `registerFilter` in namespace.

Thanks for the work here @njdehoog.

njdehoog:
You're welcome @kylef! I'm glad to see you're back on the project again. Stencil has been very valuable to my development process.

PR #48, opened 2015-12-14
Refactor error type to be a enum case

dunkelstern:
This makes error handling easier if you want to specifically catch the `TemplateParseError`

~~~swift
do {
    let t = try Template(path: templatePath)
    do {
        let body = try t.render(context, namespace: nil)
        print("Rendered template: \(body)")
    } catch StencilError.TemplateSyntaxError(let message) {
        print("Template parse error: \(message)")
    }
} catch {
    print("Could not load template \(templatePath)")
}
~~~

Further the `StencilError` enum could be expanded to send back more detailed error messages than just a generic parse error.

(I also added a Xcode project, feel free to skip that)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .conche/
 .build/
 Packages/
+xcuserdata
--- a/Sources/Include.swift
+++ b/Sources/Include.swift
@@ -8,7 +8,7 @@ public class IncludeNode : NodeType {
     let bits = token.components()
 
     guard bits.count == 2 else {
-      throw TemplateSyntaxError("'include' tag takes one argument, the template file to be included")
+      throw StencilError.TemplateSyntaxError("'include' tag takes one argument, the template file to be included")
     }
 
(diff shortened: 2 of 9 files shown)


kylef:
What's wrong with this exactly?

```swift
do {
  let body = try template.render(context)
} catch let error as TemplateSyntaxError {
  print("Template syntax error: \(error)")
}
```

Danappelxx:
Autocomplete is much more useful with enums, especially if there are a lot of errors. But since there's only one error, I'm not sure how necessary it is.

kylef:
Yes, and also there are many downsides to using an enum for this particular API:

- Any new parameters or information attached to errors require a new major release (otherwise it would break backwards compatibility), for example adding source maps to errors (#28).
- Many errors could be surfaced by plugins or custom template tags, and filters. It's not possible for those to extend the enum case.
- Any new errors require major version release.
- There's only one item so being an enum doesn't make logical sense.

PR #50, opened 2016-01-28
Fixed deprecated curried function declaration syntax

sauliusgrigaitis:


ikesyo:
This is just a function that returns a closure, not using curried function syntax from the original. Curried function syntax would be:

```swift
public func until(tags: [String])(parser: TokenParser, token: Token) -> Bool {
  if let name = token.components().first {
```

PR #64, opened 2016-03-31
Added support for swift 3.0 (DEVELOPMENT-SNAPSHOT-2016-03-24-a).

ratranqu:
Hi,

I added support for swift 3.0 (DEVELOPMENT-SNAPSHOT-2016-03-24-a) using macros checking the swift version, so should work both with 2.2 and (current) 3.0 dev.

Changes are mostly about keeping up with the swift-3-api-guidelines induced changes.

I've also updated Package.swift to point to my own updated PathKit release (see PR15 in PathKit), which you likely do not want to merge unless you do not update and release PathKit.

TODO: confirm spectre-build works with swift 3.0 (DEVELOPMENT-SNAPSHOT-2016-03-24-a) and fix where needed, as I didn't touch that at all.

### DIFF
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -17,11 +17,19 @@ public class Context {
   public subscript(key: String) -> Any? {
     /// Retrieves a variable's value, starting at the current context and going upwards
     get {
+        #if !swift(>=3.0)
       for dictionary in Array(dictionaries.reverse()) {
         if let value = dictionary[key] {
           return value
         }
-      }
+            }
+        #else
+            for dictionary in Array(dictionaries.reversed()) {
+            if let value = dictionary[key] {
+            return value
+            }
+            }
+        #endif
 
       return nil
     }
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -2,7 +2,7 @@
 public class Context {
   var dictionaries: [[String: Any]]
   let namespace: Namespace
-
+  
   /// Initialise a Context with an optional dictionary and optional namespace
   public init(dictionary: [String: Any]? = nil, namespace: Namespace = Namespace()) {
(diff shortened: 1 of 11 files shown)


Sources/Context.swift
```diff
@@ -17,11 +17,19 @@ public class Context {
   public subscript(key: String) -> Any? {
     /// Retrieves a variable's value, starting at the current context and going upwards
     get {
+        #if !swift(>=3.0)
```
kylef: This indentation is wrong, it's more indented than the life afterwards.

Sources/Context.swift
```diff
@@ -17,11 +17,19 @@ public class Context {
...
         }
-      }
+            }
+        #else
+            for dictionary in Array(dictionaries.reversed()) {
+            if let value = dictionary[key] {
+            return value
+            }
+            }
+        #endif
```
kylef: Let's break out reversing the array separately so there is less code in each branch.

```swift
#if swift(<3.0)
let dictionaries = self.dictionaries.reverse()
#else
let dictionaries = self.dictionaries.reversed()
#endif

for dictionary in dictionaries {
  ...
}
```

Sources/Parser.swift
```diff
@@ -67,15 +67,24 @@ public class TokenParser {
 
   public func nextToken() -> Token? {
     if tokens.count > 0 {
+        #if !swift(>=3.0)
```
kylef: The indentation style here is all over the place.

screen shot 2016-03-31 at 17 56 24

kylef:
Swift 3.0 support has been added in [Stencil 0.6.0](https://github.com/kylef/Stencil/releases/tag/0.6.0).

Issue #27, opened 2015-09-26
Any plan to support Carthage?


tylerlong:
https://github.com/Carthage/Carthage

I think it is easy: https://github.com/Carthage/Carthage#share-your-xcode-schemes

kylef:
The schemes are already shared, do you happen to know what else is needed to make this work with Carthage?

esttorhe:
I just tried this on my computer with the following `Cartfile`:
```swift
github "kylef/stencil"
```

And `carthage bootstrap` gives the following output:
```console
*** No Cartfile.resolved found, updating dependencies
*** Cloning Stencil
*** Checking out Stencil at "0.3.0"
*** xcodebuild output can be found in /var/folders/xk/378q_x1x4b3289cbr_kjyhzr0000gn/T/carthage-xcodebuild.sj20gi.log
*** Building scheme "Stencil" in Stencil.xcworkspace
…… <Redacted output for brevity>……
PhaseScriptExecution Copy\ Pods\ Resources /Users/<username>/Library/Developer/Xcode/DerivedData/Stencil-berzxjjyqkvunscewwpcieqtncuu/Build/Intermediates/Stencil.build/Release/StencilTests.build/Script-36A8E92EBDE6D917075028D4.sh
(33 failures)
error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.
…… <Redacted output>
/Users/<username>/Downloads/Carthage/Checkouts/Stencil/Stencil/TemplateLoader/Include.swift:2:8: error: no such module 'PathKit'
/Users/<username>/Downloads/Carthage/Checkouts/Stencil/Stencil/TemplateLoader/Include.swift:2:8: error: no such module 'PathKit'
/Users/<username>/Downloads/Carthage/Checkouts/Stencil/Stencil/TemplateLoader/Include.swift:2:8: error: no such module 'PathKit'
/Users/<username>/Downloads/Carthage/Checkouts/Stencil/Stencil/TemplateLoader/Include.swift:2:8: error: no such module 'PathKit'
/Users/<username>/Downloads/Carthage/Checkouts/Stencil/Stencil/TemplateLoader/Include.swift:2:8: error: no such module 'PathKit'
/Users/<username>/Downloads/Carthage/Checkouts/Stencil/Stencil/TemplateLoader/Include.swift:2:8: error: no such module 'PathKit'
[...]

PR #65, opened 2016-05-02
Carthage support

basememara:
Created an Xcode project, also added [PathKit](https://github.com/kylef/PathKit) as a file within the `Stencil` project because `PathKit` didn't have a dependency manager set up. Ideally though, it would be great to adjust the `Stencil` build script to pull in the `PathKit` source automatically. This PR is related to #27.

I made this [cross platform compatible with iOS, watchOS, and tvOS](http://basememara.com/creating-cross-platform-swift-frameworks-ios-watchos-tvos-via-carthage-cocoapods/). Let's wait to see next month what the best way to add macOS support ;)

Below is how the Xcode project structure was organized. Feel free to adjust as you see fit or discuss further.

capturfiles_337

It worked when I tested this in my `Cartfile`: `github "ZamzamInc/Stencil"`! 👍 

capturfiles_338

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,33 @@
 .conche/
 .build/
 Packages/
+
+## Xcode generated
+DerivedData/
+
+## Various settings
+*.pbxuser
+!default.pbxuser
+*.mode1v3
+!default.mode1v3
+*.mode2v3
+!default.mode2v3
+*.perspectivev3
+!default.perspectivev3
+xcuserdata/
+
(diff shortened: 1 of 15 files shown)


JamieWhite:
I think this should not include a copied version of PathKit. You should fix up PathKit to support Carthage. I've done just this in my fork https://github.com/skimit/pathkit which I intend on PR'ing to kylef's once I tidy it up.

PR #71, opened 2016-10-11
Templates from String Literals

vknabel:
Adds support for creating Templates from String Literals, to make dependent APIs more expressive.

```swift
func doSomething(_: Template) { ... }
doSomething("Hello {{ name }}")
```


kylef:
Thanks, this makes sense.

The test failure is unrelated and I'll fix that separately.

vknabel:
Thanks for the fast merge. 👍

Issue #76, opened 2016-12-02
Use of undeclared type 'Float80'


crsantos:
From `0.7.0` to `0.7.1`/`0.7.2` an error triggered regarding `Float80`.

```bash
Pods/Stencil/Sources/Expression.swift:205:21: Use of undeclared type 'Float80'
```

Looks like:

```swift
override func compare(lhs: Float80, rhs: Float80) -> Bool
```

Isn't `Float80` an architecture dependent type?

kylef:
Thanks for the report @crsantos. Could you please let me know which platform/architecture you are using. I wasn't aware this isn't always available.

I think we could determine the largest type available on the given platform and use that.

kylef:
Made some changes in d3706f0, but I'd like to figure out how to find the largest available number and support this. Otherwise comparisons against other largest float types would be broken.

crsantos:
Not on my computer right now but pretty sure that was running on an iPhone 6S.
On that day a quick Google search led me to this:
https://forums.developer.apple.com/thread/5026

Let me know if you need some more info when I reach home.

PR #77, opened 2016-12-07
Simple number literal parsing

djbe:
I hit this when trying to do the following:

    {% if items.count > 1 %}

Which doesn't work, because the rhs Variable evaluates to `nil`. This simply adds a number literal parser to allow this expression to work.

Small note: I'm purposefully not using a `NumberFormatter` to parse the number, because I want to ignore the user's locale for things like the decimal separator. Luckily enough, this simple way of parsing a number evaluates stuff like `2two` and `three3` to nil, so it really only parses actual numbers.


kylef:
Great! I'm 👍 on the change, I just want to get to the bottom of https://github.com/kylef/Stencil/issues/76 because this adds another use of Float80.

PR #79, opened 2016-12-07
Allow subclassing templates

kylef:
In previous versions we allowed subclassing templates but this was removed in Stencil 0.7.0 (restore in 0.7.2).

This PR brings the functionality back for the next release, while also providing an API to allow you to provide a custom template class to Environment so templates can be loaded with it.

Example Usage:

```swift
class CustomTemplate: Template {
  override func render(_ dictionary: [String: Any]? = nil) throws -> String {
    return "here"
  }
}

let environment = Environment(templateClass: CustomTemplate.self)
let result = try environment.renderTemplate(string: "Hello World")
```

**NOTE:** *Custom templates can still be used as they were, but any templates loaded by the environment can now be customised.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@
 - You can now use `{{ block.super }}` to render a super block from another `{%
   block %}`.
 
+- `Environment` allows you to provide a custom `Template` subclass, allowing
+  new template to use a specific subclass.
+
 ### Deprecations
 
 - `Template` initialisers have been deprecated in favour of using a template
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -1,9 +1,11 @@
 public struct Environment {
+  public let templateClass: Template.Type
   public let extensions: [Extension]
 
   public var loader: Loader?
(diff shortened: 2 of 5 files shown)


AliSoftware:
Wouldn't it be more logical to always ask the template to render itself (and pass the environment as a parameter to the template initializer optionally) rather than asking the environment?

Not sure if that makes sense, as I have to admit I haven't really reviewed that new "environment" concept to fully see the big picture so I may be missing something but that's just food for thoughts

kylef:
@AliSoftware I'm not sure I follow, I think that is already possible?

```swift
let template = Template(templateString: "Hello World", environment: nil)
try template.render([:])
```

The environment is a wrapper around that functionality + the template loader. Offering some connivence methods for rendering a template without handling to go create one. The template class given to the environment will be used for templates create by this convince, but also templates created by other template tags such as in the case of `{% include %}` and `{% extends %}`.

AliSoftware:
Oh, I see. You can do in one way or another!
My bad then :wink:

PR #82, opened 2016-12-12
Attach source maps to tokens and errors

kylef:
We should provide source maps so that a template author can easily understand an error raise and which part of the template cause the syntax error.

Stencil should also provide a utility to provide a HTML error page which includes the errernous component of the syntax.

As an example, providing I had the following templates:

#### `article.html`

```html+django
<h5>Comments</h5>

{% for comment in comments %}
  {% include "comment.html" %}
{% endfor %}
```

#### `comment.html`

```html+django
<h6>{{ comment.author }}</h6>

{{ comment }}

{% errblock %}
```

`{% errblock %}` throws an error inside `comment.html` included by `article.html`. The error template should contain two frames, the first frame would contain the context and source map for the `include` tag and the last frame would contain the context during `comment.html` and the source map for `err block`.

There should also be some convenience methods for printing the same information to text (stdout/stderr) perhaps with ANSI colour codes.

### Tasks

- [ ] Compute source maps for tokens.
- [ ] Attach source maps to errors.
- [ ] Ability to wrap TemplateSyntaxError for each frame.
- [ ] HTML rendering.
- [ ] Terminal and text rendering (with / without ANSI codes).

### DIFF
--- a/Sources/Lexer.swift
+++ b/Sources/Lexer.swift
@@ -1,11 +1,13 @@
 struct Lexer {
+  let fileName: String?
   let templateString: String
 
-  init(templateString: String) {
+  init(fileName: String?, templateString: String) {
+    self.fileName = fileName
     self.templateString = templateString
   }
 
-  func createToken(string:String) -> Token {
+  func createToken(string: String, sourceMap: SourceMap) -> Token {
     func strip() -> String {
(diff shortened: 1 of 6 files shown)


djbe:
@kylef Can I close this? #167 was merged in the 0.12 release.

#167 was based on #146, which was based on this PR.

PR #83, opened 2016-12-14
Normalizable.normalize() can now throw

djbe:
Let's say you have a type that is `Normalizable` but that this process can sometimes fail. This is currently not possible because the `normalize()` does not throw.

This PR simply adds this, with the necessary unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,9 @@
 - `Environment` allows you to provide a custom `Template` subclass, allowing
   new template to use a specific subclass.
 
+- `Normalizable.normalize()` now allows throwing an error should normalization
+  fail. 
+
 ### Deprecations
 
 - `Template` initialisers have been deprecated in favour of using a template
--- a/Sources/Variable.swift
+++ b/Sources/Variable.swift
@@ -69,7 +69,7 @@ public struct Variable : Equatable, Resolvable {
     }
 
     for bit in lookup() {
-      current = normalize(current)
+      current = try normalize(current)
 
       if let context = current as? Context {
         current = context[bit]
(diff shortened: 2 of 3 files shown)


AliSoftware:
Example context: https://github.com/AliSoftware/SwiftGen/pull/237#discussion_r92425077

kylef:
I don't think I understand, there is no changes that make normalise throw anything so what benefit does this have being throwable?

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@
 - `Loader`s will now throw a `TemplateDoesNotExist` error when a template
   is not found.
 
+- `Normalizable.normalize()` can now throw an error, so modify your function
```
kylef: This is an internal API so it shouldn't be mentioned in the CHANGELOG.

djbe:
I'm sorry, I forgot to make `Normalizable` public. Concrete example:

AliSoftware:
@kylef see our use case (link in my comment above) in SwiftGen for the future func/call nodes

kylef:
Would it make more sense to make use of `Resolvable` (or perhaps `NodeType`) instead https://github.com/kylef/Stencil/blob/fdde1dec029b25198e34129f10dc8dbc51028b5f/Sources/Node.swift#L55-L57 ?

```swift
public protocol Resolvable {
  func resolve(_ context: Context) throws -> Any?
}

public protocol NodeType {
  /// Render the node in the given context
  func render(_ context:Context) throws -> String
}
```

`normalize` is really more designed to work around some Swift typing problems to normalise types (standardise arrays and dicts etc). It doesn't seem like you want to use normalise to normalise types but instead resolve/render them in a specific way so I think some of the existing types may be better suited for your use.

djbe:
The thing is that we have 2 node types already: `FuncNode` and `CallNode`. The first one generates the `CallableBlock` that is the issue here, which is then stored in the context as a `Variable`. 

Anyway, I now realise that I jumped the gun on this, and it is indeed not needed. Apologies 😊

PR #85, opened 2016-12-17
Trim newlines

### LINKED ISSUE #22: Lots of blank empty lines in output

yonaskolb:
This fixes #22 and is an alternative to #32 
The lexer removes white space and trailing newlines from block tokens if they are on their own line.
```
{% for item in items %}
- {{item.name}}
{% endfor %}
```
will now be rendered
```
- item 1
- item 2
- item 3
```

### DIFF
--- a/Sources/Lexer.swift
+++ b/Sources/Lexer.swift
@@ -50,7 +50,29 @@ struct Lexer {
       }
     }
 
-    return tokens
+    var trimmedTokens = tokens
+    for (index, token) in tokens.enumerated() {
+        if case .block = token, tokens.count > 1 {
+            let trimPattern = "^[\t ]*\n"
+            if index < tokens.count - 1,
+                case .text(let nextText) = tokens[index + 1],
+                nextText.range(of: trimPattern, options: .regularExpression, range: nil, locale: nil) != nil {
+                if index == 0 {
+                    let trimmedNextText = trimmedTokens[index + 1].contents.replacingOccurrences(of: trimPattern, with: "", options: .regularExpression, range: nil)
+                    trimmedTokens[index + 1] = .text(value: trimmedNextText)
+                }
+                else if case .text(let previousText) = tokens[index - 1],
+                    previousText.range(of: "\n[\t ]*$", options: .regularExpression, range: nil, locale: nil) != nil,
+                    !(index == tokens.count - 2 && nextText == "\n") {
+                    let trimmedPreviousText = trimmedTokens[index - 1].contents.replacingOccurrences(of: "[\t ]*$", with: "", options: .regularExpression, range: nil)
+                    let trimmedNextText = trimmedTokens[index + 1].contents.replacingOccurrences(of: trimPattern, with: "", options: .regularExpression, range: nil)
+                    trimmedTokens[index - 1] = .text(value: trimmedPreviousText)
+                    trimmedTokens[index + 1] = .text(value: trimmedNextText)
+                }
+            }
+        }
+    }
+    return trimmedTokens
   }
 }
 
--- a/Sources/Lexer.swift
+++ b/Sources/Lexer.swift
@@ -62,8 +62,7 @@ struct Lexer {
                     trimmedTokens[index + 1] = .text(value: trimmedNextText)
                 }
                 else if case .text(let previousText) = tokens[index - 1],
-                    previousText.range(of: "\n[\t ]*$", options: .regularExpression, range: nil, locale: nil) != nil,
-                    !(index == tokens.count - 2 && nextText == "\n") {
+                    previousText.range(of: "\n[\t ]*$", options: .regularExpression, range: nil, locale: nil) != nil {
                     let trimmedPreviousText = trimmedTokens[index - 1].contents.replacingOccurrences(of: "[\t ]*$", with: "", options: .regularExpression, range: nil)
                     let trimmedNextText = trimmedTokens[index + 1].contents.replacingOccurrences(of: trimPattern, with: "", options: .regularExpression, range: nil)
                     trimmedTokens[index - 1] = .text(value: trimmedPreviousText)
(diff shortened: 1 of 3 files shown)


yonaskolb:
```
!(index == tokens.count - 2 && nextText == "\n")
```
At line 66, is only there to keep 2 existing tests passing which expected 2 trailing newlines at the end of the file. This could be removed and the tests changed to maintain consistent behaviour

yonaskolb:
I've gone ahead and removed the special case for end of files newlines and fixed those 2 tests

yonaskolb:
Any comments @kylef ?

AliSoftware:
@yonaskolb I think @kylef has already stated in multiple issues and PR that he rather wanted to match what ejs and Jinja2 are using, making it explicit to not want the newline using a `-` prefix for example.

See https://github.com/kylef/Stencil/issues/89#issuecomment-270760145

kylef:
That is mostly correct, however I haven't given this sufficient thought and gone over all the user-cases. It could perhaps make sense to offer default options inside a Stencil environment, for example `Environment(trimBlocks: true)` or `Environment(leftStripBlocks: true)`.

Where the default behaviour may very well be stripping whitespace/newlines by default, providing the user can disable this behaviour either at the render level (environment) or perhaps at the tag level.

kylef:
I'd also like to explore the performance implications of these changes as with basic initial testing these changes would make Stencil 25.49% slower. I don't this change should be default behaviour if it effects the performance so much. Especially given these changes would only effects formatting of rendered templates, which for most uses these will rarely be seen by humans and just by web browsers.

- Master (99efba5): 1000 iterations is 4.52544301748276
- PR (ffd8804): 1000 iterations is 5.67912900447845

Here's the template and code I used to get these numbers (adapted from a Django perf test):

```html+django
<html>
<body>
    <div id="content">
    {% if items %}
    <ul>
        <li>Single item:</li>
        <li>
        {{ item.0.name }}
        </li>
    {% for item in items %}
    
        {% if item.name %}
        <li>{{ item.name }}</li>
        {% endif %}
        {% if not item.name %}
        <li>No name item</li>
        {% endif %}
        {% for item in fooitems %}
        <li>{{ item.name }}</li>
        {% endfor %}
        {% if item.des %}
        <li>{{ item.des }}</li>
        {% endif %}
    {% endfor %}
    {% if items %}
        {% for item in items %}
        
        <li>{{ item.name }}</li>
        {% endfor %}
    {% endif %}
    </ul>
    {% endif %}
    {% if items %}
    <ul>
    {% for item in items %}
        {{ item.name }}
    {% endfor %}
    {% for foo in foos %}
    
        {{ foo }}
    {% endfor %}
    {% for item in items %}
        <li>huh{{ foo }}</li>
    {% endfor %}
    </ul>
    {% endif %}
    {% if items %}<p>We have items.</p>{% endif %}
    {% if items %}
    <p>We have items.</p>
    {% endif %}
    {% if foo %}<p>We have foo.</p>{% endif %}
    {% if foo %}
    <p>We have foo.</p>
    {% endif %}
    </div>
</body>
</html>
```

```swift
import func CoreFoundation.CFAbsoluteTimeGetCurrent
import Stencil
import PathKit

func time(closure:(() -> ())) -> Double {
  let start = CFAbsoluteTimeGetCurrent()
  closure()
  let end = CFAbsoluteTimeGetCurrent()

  return end - start
}

struct Item {
  var name: String
}

let items = (0 ..< 1).map { index in
  Item(name: "Item \(index)")
}
let context = [
  "items": items
]

let templateString = try Path("example.html").read() as String

let total = time {
  (0..<1000).map { _ in
      _ = try! Template(templateString: templateString).render(context)
  }
}

print("Total time taken for 1000 iterations is \(total).")
```

yonaskolb:
Yeah I think having it as an option on environment is a good idea. I can make that change in this pull request. As far as performance goes, I guess the regular expressions are slowing things down. Would some sort of scanner be faster?
In terms of tags, what sort of block tags would you not want this on?

Trimming these lines is very important for my use case which is generating use facing code. Having the blocks on newlines makes reading and editing the templates so much easier, and because it's consumed by humans the formatting is important. 
Here's an example template which is utilising this branch
https://github.com/yonaskolb/SwagGen/blob/master/Templates/Swift/model.swift

kylef:
@yonaskolb I apologise that the whole topic of trimming new lines has been going on for so long. I'd like to get the behaviour settled so we can move forward. I do like the alternative proposed in #92, as it gives the template author control of the behaviour and ability to override the default behaviour on per use basis.

I would consider extending on #92 to add the ability to supply the default trimming behaviour (without `-` or `+`), but I am concerned that it complicates template authoring to the point where templates have different behaviour based on some configuration. When moving templates from one project to another the behaviour significantly changes. This however may not be a problem in practise, just a 💭.

Do you think #92 would work?

/c @AliSoftware

AliSoftware:
Using `{%- tag -%}` to do trimming like in #92 would totally work for me. It's kind of a standard in many templating languages anyway and brings all the necessary flexibility 👍

AliSoftware:
And I agree that giving the ability to configure the default behavior is probably not worth it. Fwiw I think not trimming could be the default and `-` would be enough to activate the trimming on demand

yonaskolb:
Yes #92 looks like a good solution. 
In my case I _would_ want to see the ability to specify an environment's default behaviour. In my opinion if you have a load of templates that are single domain specific (like a huge group of files that will be turned into human readable source code) it keeps the templates much lighter without specifying `-` everywhere.

I can appreciate the sentiment that it would make the templates not as transferrable, but with custom filters and all the extensions people are writing for Stencil I think that ship might have sailed anyway. Plus from the use cases I've seen there haven't been any templates distributed without the code to generate them.

yonaskolb:
Closing in favour of #92

PR #86, opened 2016-12-20
Basic numeric expresion handling

djbe:
Stencil has boolean expression handling, but numeric ones. Instead of writing my own ridiculously  complex expression parser, there exists this little gem of a Foundation class called `NSExpression`, so I just used that.

Small problem is that, while it accepts an `object` parameter, this is quite limited, as it won't know how to handle `myArray.count` because `myArray` isn't Key-Value compliant for `count`, and all object leaves must be number values. That's why I rolled my own resolver using Stencil's existing system.

There is a small gotcha with this whole system: if `NSExpression` fails parsing an expression, it throws an Objective-C exception, which Swift can't catch. There are solutions for this, such as http://stackoverflow.com/questions/32758811/catching-nsexception-in-swift, but they would require a small bit of objective-c code.

I've just made a quick implementation to throw the idea out there. It currently works for examples such as `{{ (myArray.count + 5) / 3 - 1 }}`, which is already really cool.

### DIFF
--- a/Sources/Variable.swift
+++ b/Sources/Variable.swift
@@ -3,6 +3,14 @@ import Foundation
 
 typealias Number = Float
 
+public protocol GenericVariable: Equatable {
+  var variable: String { get }
+}
+
+public func ==<T: GenericVariable>(lhs: T, rhs: T) -> Bool {
+  return lhs.variable == rhs.variable
+}
+
 
 class FilterExpression : Resolvable {
   let filters: [(FilterType, [Variable])]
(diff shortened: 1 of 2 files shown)


krzysztofzablocki:
fwiw. I use catching objective-c exceptions in Sourcery to recover from Stencil KVC resolution failures (in daemon mode)

djbe:
You use this, right?
https://github.com/williamFalcon/SwiftTryCatch

krzysztofzablocki:
Yeah a fork

kylef:
@djbe not having much luck with testing this. Am I doing something wrong?

```
 CompoundVariable can resolve a literal expression
  Tests/StencilTests/VariableSpec.swift:125 given value is nil

  ``
  try expect(result) == "3"
  ``
 CompoundVariable can resolve a variable expression
  Tests/StencilTests/VariableSpec.swift:131 given value is nil

  ``
  try expect(result) == "15"
  ``
```

```diff
diff --git a/Tests/StencilTests/VariableSpec.swift b/Tests/StencilTests/VariableSpec.swift
index d66dc1a..0723c40 100644
--- a/Tests/StencilTests/VariableSpec.swift
+++ b/Tests/StencilTests/VariableSpec.swift
@@ -19,16 +19,18 @@ fileprivate struct Article {
 
 
 func testVariable() {
-  describe("Variable") {
-    let context = Context(dictionary: [
-      "name": "Kyle",
-      "contacts": ["Katie", "Carlton"],
-      "profiles": [
-        "github": "kylef",
-      ],
-      "article": Article(author: Person(name: "Kyle"))
-    ])
+  let context = Context(dictionary: [
+    "name": "Kyle",
+    "contacts": ["Katie", "Carlton"],
+    "profiles": [
+      "github": "kylef",
+    ],
+    "article": Article(author: Person(name: "Kyle")),
+    "x": 5,
+    "y": 10,
+  ])
 
+  describe("Variable") {
 #if os(OSX)
     context["object"] = Object()
 #endif
@@ -115,4 +117,18 @@ func testVariable() {
     }
 #endif
   }
+
+  describe("CompoundVariable") {
+    $0.it("can resolve a literal expression") {
+      let variable = CompoundVariable("1 + 2")
+      let result = try variable.resolve(context) as? String
+      try expect(result) == "3"
+    }
+
+    $0.it("can resolve a variable expression") {
+      let variable = CompoundVariable("x + y")
+      let result = try variable.resolve(context) as? String
+      try expect(result) == "15"
+    }
+  }
 }
```

djbe:
@kylef I've added some tests that should show how it works. Small issue with your test code is that the result of numeric expressions will be a number, not a string.

DivineDominion:
Would like to hereby +1 this :) 👍

ilyapuchka:
@djbe can we close this in favour of #183

djbe:
I don't know, I'm still not sure which is better.

- `NSExpression` is much, much more flexible in what it allows a user to do, we get all the operators for "free", but doesn't work on Linux yet.
- With #183 we have Linux support, but only very limited expression support. Every other operator we may want to add, we'll have to write, test and maintain ourselves.

ilyapuchka:
I feel Linux support is a better thing to have for project like Stencil than having a very powerful expressions engine under the hood 🤷‍♂️

djbe:
If we ever implement this, we should use a library such as [Expression](https://github.com/nicklockwood/Expression).

The mechanism for triggering an expression "parse" could be taken from this PR, and we'd need to provide expression with our own symbol lookup function (which it supports).

PR #88, opened 2016-12-25
[WIP] Add Carthage Support

RahulKatariya:


RahulKatariya:
Let me know your thoughts.

kylef:
Hi @RahulKatariya, I appreciate your efforts but I do not want to use Carthage as a dependency manager for Stencil.

Carthage unfortunately forces library authors to use Carthage to manage a libraries dependencies to support Carthage in their recommended way. This also forces Stencil contributors to use Carthage to manage and update dependencies of Stencil in the future, which inturn forces you to develop Stencil on macOS.

I've provided some steps on how you can use Stencil with Carthage in the documentation for Stencil at https://stencil.fuller.li/en/latest/installation.html#carthage by using the generated Xcode project from `swift package generate-xcodeproj`.

We do also offer support for CocoaPods, but that is purely because CocoaPods doesn't force us to change anything in the library or forces the library to be developed in a specific way. Instead as [one of their goals](https://github.com/CocoaPods/CocoaPods#project-goals) they allow library authors to develop their projects however they want and allow users to integrate libraries without the library author doing anything specific for CocoaPods.

PR #96, opened 2017-01-22
Support where expression in for tag

ilyapuchka:
This pr adds support for `where` expression in `for` tag similar to Swift `where`. With it inner context of the loop will receive already filtered results.

```
{% for type in types.all where type.name|hasPrefix:\"Foo\" %}
{{ forloop.counter }}: {{ type.name }}\n
{% endfor %}
```

Currently it's possible to achieve somewhat similar result using `if` tag inside `for` tag:

```
{% for type in types.all %}
{% if type.name|hasPrefix:\"Foo\" %}
{{ forloop.counter }}: {{ type.name }}\n
{% endif %}
{% endfor %}
```

But in this case the loop will receive not filtered collection, so loop context properties (counter, first, last) will be set based on non-filtered collection. With `where` they will be set based on filtered collection, so it will be possible i.e. to correctly compare current item with first or last items, etc.

@kylef please tell me what you think about this and if your are ok with merging it I will go on with wrapping it by adding tests etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Stencil Changelog
 
+## Master
+
+- For expression now can contain `where` expression to filter array items. For example `{% for item in items where item > 1 %}` is now supported.
+
 ## 0.8.0
 
 ### Breaking
--- a/Sources/ForTag.swift
+++ b/Sources/ForTag.swift
@@ -3,12 +3,14 @@ class ForNode : NodeType {
   let loopVariable:String
   let nodes:[NodeType]
   let emptyNodes: [NodeType]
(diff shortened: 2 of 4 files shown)


kylef:
@ilyapuchka this looks great to me, would you be able to add some test coverage for loops using `where` along with a CHANGELOG entry and I'll merge this up. Thanks for the work here!

ilyapuchka:
@kylef I've added some tests and added some docs, please let me know if you want to add more test cases

ilyapuchka:
@kylef can you please take a look at this pr again, when you have some time? =) I really hope this will make it to the next release with `elif` support =)

Issue #94, opened 2017-01-18
Support iterating over dictionaries


kylef:
```swift
{% for key, value in thing %}
    <li>{{ key }}: {{ value }}</li>
{% endfor %}
```

tprasad23:
Hi, I received an error when attempting this, 

`Error raised at top level: 'for' statements should use the following 'for x in y' `for key, value in categories`
`
any suggestions?

PR #98, opened 2017-01-24
Fixed parsing several filter arguments with quotes

ilyapuchka:
Current version (latest master) incorrectly parses subsequent filter parameters that use quotes, i.e. `name|replace:\"foo\",\"bar\"` will give three arguments instead of 2.

Internally it creates filter expression with three variables, `"foo"`, `":"` and `"bar"`. It inserts unneeded colon because when splitting filter token by colon it splits it into `replace`, `"foo"` and `,"bar"`, then it drops the first and joins all others with colon. That gives arguments string `"foo":,"bar"`. Then it also incorrectly splits it into `"foo"`, `:` and `"bar"`.

With this fix token is correctly split  into `replace`, `"foo","bar"` and then arguments are split into `"foo"` and `"bar"`

### DIFF
--- a/Sources/Tokenizer.swift
+++ b/Sources/Tokenizer.swift
@@ -7,14 +7,18 @@ extension String {
     var word = ""
     var components: [String] = []
     var separate: Character = separator
+    var singleQuoteCount = 0
+    var doubleQuoteCount = 0
 
     for character in self.characters {
+      if character == "'" { singleQuoteCount += 1 }
+      else if character == "\"" { doubleQuoteCount += 1 }
+
       if character == separate {
+
         if separate != separator {
           word.append(separate)
-        }
-
-        if !word.isEmpty {
+        } else if singleQuoteCount % 2 == 0 && doubleQuoteCount % 2 == 0 && !word.isEmpty {
           components.append(word)
           word = ""
         }
--- a/Tests/StencilTests/FilterSpec.swift
+++ b/Tests/StencilTests/FilterSpec.swift
@@ -22,8 +22,8 @@ func testFilter() {
       try expect(result) == "Kyle Kyle"
     }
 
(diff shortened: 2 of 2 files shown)


Sources/Tokenizer.swift
```diff
@@ -7,18 +7,22 @@ extension String {
     var word = ""
     var components: [String] = []
     var separate: Character = separator
+    var singleQuoteCount = 0
+    var doubleQuteCount = 0
```
kylef: Type: `Qute`

Tests/StencilTests/FilterSpec.swift
```diff
@@ -35,7 +35,23 @@ func testFilter() {
       }
 
       let result = try template.render(Context(dictionary: context, environment: Environment(extensions: [repeatExtension])))
-      try expect(result) == "Kyle Kyle with args value"
+      try expect(result) == "Kyle Kyle with args value1, \"value2\""
+    }
+    
+    $0.it("allows you to register a custom filter which accepts several arguments") {
+        let template = Template(templateString: "{{ name|repeat:'value\"1\"',\"value'2'\",'(key, value)' }}")
+
```
kylef: Theres lots of trailing whitespace here, could you please remove / configure editor not to put these here?

ilyapuchka:
@kylef done

ilyapuchka:
@kylef can you please take another looks at this pr?

kylef:
@ilyapuchka Yes, sorry for delay in getting back to you on some of these PRs. Struggling to find time lately for OSS.

I really appreciate all the time and effort you've put in here, I'll make a release shortly with your changes.

kylef:
I'll try tackling https://github.com/kylef/Stencil/issues/94 tomorrow and then make a release.

PR #106, opened 2017-04-03
Fixed warnings with xcode 8.3.

avinashdongarwar:
This PR addresses the warnings in following screenshot. Kindly lemme know for any other expected change.
screen shot 2017-04-03 at 1 00 15 pm

### DIFF
--- a/Sources/Template.swift
+++ b/Sources/Template.swift
@@ -14,7 +14,7 @@ open class Template: ExpressibleByStringLiteral {
   public let name: String?
 
   /// Create a template with a template string
-  public required init(templateString: String, environment: Environment? = nil, name: String? = nil) {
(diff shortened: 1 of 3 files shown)


Sources/Template.swift
```diff
@@ -14,7 +14,7 @@ open class Template: ExpressibleByStringLiteral {
   public let name: String?
 
   /// Create a template with a template string
-  public required init(templateString: String, environment: Environment? = nil, name: String? = nil) {
+  public required init(templateString: String, environment: Environment? = nil, name: String? = nil) throws {
```
kylef: There is no case this would ever throw, why should it be annotated as throws?

avinashdongarwar: @kylef If so, why do you `try` it on Line no.55 and 68 in `Loader.swift`? It made me believe that you are doing it intentionally and i ended up making this throwable.

PR #110, opened 2017-04-28
Handle double optionals

tomcomer:
When using Stencil with Vapor, I found that the id field of model objects would resolve to a double optional within Stencil. The id field is defined as Node? and the current variable here is defined as Any?, so what it ends up being is Optional(Optional(Node)). The code I added below handles that case.

### DIFF
--- a/Sources/Variable.swift
+++ b/Sources/Variable.swift
@@ -99,9 +99,15 @@ public struct Variable : Equatable, Resolvable {
         let mirror = Mirror(reflecting: value)
         current = mirror.descendant(bit)
 
-        if current == nil {
-          return nil
-        }
+        // handle the possiblity of double optionals
+        if let obj = current {
+            let r = Mirror(reflecting: obj)
+            if let display = r.displayStyle, display == .optional {
+                current = r.descendant("some")
+            }
+        } else {
+            return nil
+        }        
       } else {
         return nil
       }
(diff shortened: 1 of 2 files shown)


kylef:
Hi @tomcomer, thanks for your contribution. Would you be able to add a test-case to demonstrate this behaviour and to prevent any further regressions in the future.

tomcomer:
I've added a test-case and updated the PR.

kylef:
This test passes without other sources changes. Either this is already fixed, or this isn't a real bug or this test isn't testing the desired behaviour:

```shell
$ hub fetch tomcomer
$ git checkout tomcomer/master Tests/StencilTests/VariableSpec.swift
$ swift test
```

I'm going to close, please feel free to create a new pull request if you can demonstrate there is a bug with a test that fails on master.

djbe:
@kylef This issue is still present for me. Say I have the following data:
```swift
final class Entity {
    let name: String
    weak var parent: Entity?
}

let context = [
  "entities": [... a bunch of entities here...]
]
```

And I try the following template:
```
{% for entity in entities %}
  {{ entity.name }}: {{ entity.parent.name }}
{% endfor %}
```

It will never reach the `name` property of a parent, because:
- `mirror.descendant("parent")` returns a "Optional<Optional< Entity>>"
- for the next `bit` iteration, it'll try the "name" property
- it first unwraps `if let value = current` --> `value` is now "Optional< Entity>"
- `mirror.descendant("name")` on that optional results in nil

Funnily enough, the following template **does** work:
```
{% for entity in entities %}
  {{ entity.name }}: {{ entity.parent.some.name }}
{% endfor %}
```
But that's not very intuitive for users.

ilyapuchka:
@djbe I think #204 should fix this

PR #112, opened 2017-05-02
Dictionary tweaks

yonaskolb:
This allows you to access the count of a dictionary. If the dictionary happens to have a value at the key "count" it will return that value instead. Includes tests

I also noticed that the dictionary for loop doesn't support a space after the comma, so that this is supported:
```
{% for key,value in thing %}
```
but this is not:
```
{% for key, value in thing %}
```
This just updates the changelog to reflect this discrepancy until it is fixed. Once it is, the `can iterate over dictionary` test can have a space added

### DIFF
--- a/Sources/Variable.swift
+++ b/Sources/Variable.swift
@@ -74,7 +74,11 @@ public struct Variable : Equatable, Resolvable {
       if let context = current as? Context {
         current = context[bit]
       } else if let dictionary = current as? [String: Any] {
-        current = dictionary[bit]
+        if bit == "count" && dictionary[bit] == nil {
+          current = dictionary.count
+        } else {
+          current = dictionary[bit]
+        }
       } else if let array = current as? [Any] {
         if let index = Int(bit) {
           if index >= 0 && index < array.count {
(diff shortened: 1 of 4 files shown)


Sources/Variable.swift
```diff
@@ -74,7 +74,11 @@ public struct Variable : Equatable, Resolvable {
       if let context = current as? Context {
         current = context[bit]
       } else if let dictionary = current as? [String: Any] {
-        current = dictionary[bit]
+        if bit == "count" && dictionary[bit] == nil {
```
kylef: I am not too sure about this behaviour and worried it could lead to hard to debug problems for template authors when they use `count`. Sometimes `count` would be the number of items in a dictionary, but some other times it could be an item from the dictionary leading to confusion.

yonaskolb: Yes, there could potentially be a conflict there. 
What do you suggest the fix is? Always read `count` as the count of the dictionary? Rename it to a less likely name like `dictionaryCount`, but then lose consistency with array?

kylef: I think we should go with `count` always, it would always be the count of items in the collection. Then it will be consistent with array, and its going to be more likely you want to get the count of the sequence over the value of count.

kylef:
Merged this in as 14bac03 (with the suggested change) and bf4be38.

PR #114, opened 2017-05-22
[Review] Include block improvements

### LINKED ISSUE #95: Question: Indentation control in {% include %}
Hi. I'm trying to use stencil for swift code generation (as many other existing applications of stencil) and found it very confusing when it comes to indentation control. It isn't very important for html code generation (as  jinja, liquid, django and many other implementations if this kind of templating usually do) because that html is then being interrupted by browsers and humans don't look in that geneted html. But for swift code generation it would be beneficial for generated code too look readable for humans too.

Simple example to describe the problem. Consider template
```
struct SomeType {
    {% include variablesSection%}
    {% include publicMethosSection %}
}
```

and then `publicMethosSection` too look like 
```
func method1() {
    doSomething()
}

func method2() {
    doSomething()
}
```

Master template assumes some indentation as both includes are prefixed with 4 spaces, however when rendering includes inner templates all new lines generated by inner template break that indentation.

In terms of templating language design, what is the suitable way of ensuring that all new lines coming from include gets automatically indented same way as `include` clause in master template? I understand this feature  may not exist yet, but how you think it may look like?

yonaskolb:
Fixes #95

This adds 2 features to the `include` block including tests:
1. If the include block is indented with any whitespace, any content in the included file is indented by the same amount
2. Optional argument `using` which lets you pass a context to the included file using the format: `{% include "myFile.stencil" using myContext %}`

### DIFF
--- a/Sources/Parser.swift
+++ b/Sources/Parser.swift
@@ -56,7 +56,7 @@ public class TokenParser {
           if nodes.count > 1,
             let indentedNode = nodes[nodes.count - 1] as? Indented,
             let textNode = nodes[nodes.count - 2] as? TextNode {
-            indentedNode.indent = textNode.text.trimmingCharacters(in: CharacterSet.whitespaces.inverted)
+            indentedNode.indent = textNode.text.components(separatedBy: "\n").last!.trimmingCharacters(in: CharacterSet.whitespaces.inverted)
           }
         }
       case .comment:
--- a/Sources/Parser.swift
+++ b/Sources/Parser.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 public func until(_ tags: [String]) -> ((TokenParser, Token) -> Bool) {
   return { parser, token in
     if let name = token.components().first {
@@ -50,6 +52,12 @@ public class TokenParser {
         if let tag = token.components().first {
           let parser = try findTag(name: tag)
           nodes.append(try parser(self, token))
+
+          if nodes.count > 1,
+            let indentedNode = nodes[nodes.count - 1] as? Indented,
+            let textNode = nodes[nodes.count - 2] as? TextNode {
+            indentedNode.indent = textNode.text.trimmingCharacters(in: CharacterSet.whitespaces.inverted)
+          }
         }
       case .comment:
         continue
@@ -94,5 +102,8 @@ public class TokenParser {
   public func compileFilter(_ token: String) throws -> Resolvable {
     return try FilterExpression(token: token, parser: self)
   }
+}
 
+protocol Indented: class {
+  var indent: String { get set }
 }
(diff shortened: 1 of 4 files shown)


yonaskolb:
Any comments on this @kylef?

ilyapuchka:
+1

ilyapuchka:
@kylef this looks good to me and feels like very useful fix. Any chance you can look into it any time soon or you ok with me merging it? =)

yonaskolb:
Any updates on this?

Sources/Parser.swift
```diff
@@ -50,6 +52,12 @@ public class TokenParser {
         if let tag = token.components().first {
           let parser = try findTag(name: tag)
           nodes.append(try parser(self, token))
+
+          if nodes.count > 1,
+            let indentedNode = nodes[nodes.count - 1] as? Indented,
+            let textNode = nodes[nodes.count - 2] as? TextNode {
+            indentedNode.indent = textNode.text.components(separatedBy: "\n").last!.trimmingCharacters(in: CharacterSet.whitespaces.inverted)
```
ilyapuchka: I was looking into this and I think there might be an issue with this implementation. What if include tag line is prefixed with something? i.e.

```
Something
  Something:  {% include \"include.html\" %}
```

If I understand the logic here you are getting the last line of the text preceding include tag, so this will be `"  Something:  "`. And then you are trimming everything which is _not_ a whitespace. In this case nothing will be trimmed, so you will write to the `indent` the whole string, which will then add it to each line of included content. Right?

Also what do you think should be behaviour for indentation in this case. Should the second line of included content be indented at all? And if it should then how much? To match the offset of first line of included text or to match the offset of previous line in template? I.e. it should be 

```
Something
  Something:  first line
              second line
```

or it should be 

```
Something
  Something:  first line
  second line
```
? Should we give user an option to control this?

Also there is [this pr](#92) that touches indentation too, so that if you do `{%+ include ... %}` it should keep spaces before tag. So this might conflict with each other.

I would therefore extract another feature that you've implemented here, for passing sub context to included template, into separate PR and maybe rethink indentation implementation. Maybe it should be done only after #92 

@kylef what do you think?

ilyapuchka: Maybe alternative solution for this could be `indent` node that will wrap included content (or any other content, i.e. result of macro call) and will define how all the content lines should be indented. Jinja2 has `indent` filter for that, but that will require to render content into variable to which the filter can be then applied, which is not yet supported by Stencil (though it is implemented in [StencilSwiftKit](https://github.com/ilyapuchka/StencilSwiftKit/blob/master/Sources/SetNode.swift)). I think we can do both, node and filter to cover all scenarios, similar to how we have `filter` node in addition to filters applicable to variables.

@kylef thoughts?

ilyapuchka: Here is what I ended up with #186 , there was no need for a node, only for a filter, thanks to `filter` tag.

yonaskolb:
Closing this in favour of #188 in terms of indenting.

I might open a new PR for passing the context however. It seems the direction recently is to go with unnamed arguments, is that preferred to `using`?

PR #123, opened 2017-07-11
Added example to show use of `forloop.x`in a for loop

iandundas:
Was not immediately obvious to me :)


kylef:
Thanks @iandundas

PR #130, opened 2017-08-10
Fix example code in getting started

paulmelnikow:
Thanks for this useful library!

This bit of example code confused me so I wanted to correct it. Cheers!


kylef:
Thanks @paulmelnikow

PR #134, opened 2017-08-30
Make Template render(_ context: Context) public

AttilaTheFun:
I'd like to make this function public so I can call it from my library that consumes Stencil. It's necessary for a custom node I implemented.

### DIFF
--- a/Sources/Template.swift
+++ b/Sources/Template.swift
@@ -63,7 +63,7 @@ open class Template: ExpressibleByStringLiteral {
   }
 
   /// Render the given template with a context
-  func render(_ context: Context) throws -> String {
+  public func render(_ context: Context) throws -> String {
     let context = context
     let parser = TokenParser(tokens: tokens, environment: context.environment)
     let nodes = try parser.parse()


kylef:
The extension point is to create an Extension and register it with your Environment. (https://stencil.fuller.li/en/latest/custom-template-tags-and-filters.html). For this you do not need this method to be public.

AttilaTheFun:
Hi @kylef, thanks for your reply. I should have been more specific. My library depends on [SwiftStencilKit](https://github.com/SwiftGen/StencilSwiftKit) and your library. Specifically, I make use of their macro tags, which allow you to define lightweight functions in your stencil files. While I implement most functions as custom filters, I use macros for language/implementation specific logic that is repeated in multiple templates. One of the frustrating limitations was that they didn't play nicely with the include tag. If I wanted to define them in a separate file, the include rendered everything in that file, and then popped its context off the stack. I wanted to do the opposite - DON'T render the other file, but push all of its macros onto the stack, and then they're available while my custom tag is on the stack. Here is my ImportNode:

```
class ImportNode : NodeType {
    let templateName: Variable
    let nodes: [NodeType]

    class func parse(_ parser: TokenParser, token: Token) throws -> NodeType {
        let bits = token.components()

        guard bits.count == 2 else {
            throw TemplateSyntaxError("'import' tag takes one argument, the template file to be included")
        }

        let scopedNodes = try parser.parse(until(["endimport"]))

        guard parser.nextToken() != nil else {
            throw TemplateSyntaxError("`endimport` was not found.")
        }

        return ImportNode(templateName: Variable(bits[1]), nodes: scopedNodes)
    }

    init(templateName: Variable, nodes: [NodeType]) {
        self.templateName = templateName
        self.nodes = nodes
    }

    func render(_ context: Stencil.Context) throws -> String {
        guard let templateName = try self.templateName.resolve(context) as? String else {
            throw TemplateSyntaxError("'\(self.templateName)' could not be resolved as a string")
        }

        let template = try context.environment.loadTemplate(name: templateName)
        
        return try context.push {
            let _ = try template.render(context)
            let renderedNodes = try renderNodes(nodes, context)
            return renderedNodes
        }
    }
}
```

PR #144, opened 2017-09-30
fixed checking values for nil when using mirror

### LINKED ISSUE #140: Context includes optionals holding `nil`
I am using Stencil with Kitura and noticed that, when building a rendering context with instances of my own types (instead of a dictionary), that context includes properties that are `nil`. The `if` tag evaluates these properties as `true`, which is very confusing.

For example:
```
{% if post.lastUpdate %}
Last updated on {{ post.lastUpdate }}
{% endif %}
```
results in
```
Last updated on
```
because the `lastUpdate` property is present in the context even if it is `nil`.

ilyapuchka:
Resolves #140 
@kylef couldn't find a simpler way to check for nil underlying values. There is a way to do that using custom box protocol, but it requires extending both Option and IUO types, but string approach works for both of them already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   incorrect amount of arguments.
 - Fixes a potential crash when using incomplete tokens in a template for
   example, `{%%}` or `{{}}`.
+- Fixes evaluating nil properties as true
 
 
 ## 0.9.0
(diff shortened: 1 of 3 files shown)


kylef:
Hi @ilyapuchka, thanks for looking into this. I guess this is the best way for now, would you like to add a CHANGELOG entry for this bug fix and then I'll merge it.

ilyapuchka:
👍 done

PR #146, opened 2017-10-03
Improved error reporting

ilyapuchka:
Added file name, line number and failed token highlighting in error message

Old:

```
`endfor` was not found.
```

New:

```
TemplateName:1:0: error: `endfor` was not found.
{% for name in names %}{{ name }}
^~~~~~~~~~~~~~~~~~~~~~~
```

@kylef this is an attempt to solve #82 to provide better error messages. I've added some tests to demonstrate new error messages. Let me know if you like this approach and if I should proceed. I plan to use similar approach to filters errors and other things (though they will require some other changes along the way).

### DIFF
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -47,7 +47,7 @@ public struct Environment {
     do {
       return try template.render(context)
     } catch {
-      try errorReporter.report(error: error)
+      throw errorReporter.reportError(error)
     }
   }
   
@@ -66,7 +66,7 @@ public struct Environment {
     do {
       return try closure()
     } catch {
-      try errorReporter.report(error: error)
+      throw errorReporter.reportError(error)
     }
   }
   
--- a/Sources/Errors.swift
+++ b/Sources/Errors.swift
@@ -46,17 +46,17 @@ public class ErrorReporterContext {
 
 public protocol ErrorReporter: class {
   var context: ErrorReporterContext! { get set }
-  func report(error: Error) throws -> Never
+  func reportError(_ error: Error) -> Error
   func contextAwareError(_ error: TemplateSyntaxError, context: ErrorReporterContext) -> Error?
 }
 
 open class SimpleErrorReporter: ErrorReporter {
   public var context: ErrorReporterContext!
   
-  open func report(error: Error) throws -> Never {
-    guard let syntaxError = error as? TemplateSyntaxError else { throw error }
-    guard let context = context else { throw error }
-    throw contextAwareError(syntaxError, context: context) ?? error
+  open func reportError(_ error: Error) -> Error {
+    guard let syntaxError = error as? TemplateSyntaxError else { return error }
+    guard let context = context else { return error }
+    return contextAwareError(syntaxError, context: context) ?? error
   }
   
   // TODO: add stack trace using parent context


Sources/Node.swift
```diff
@@ -3,10 +3,22 @@ import Foundation
 
 public struct TemplateSyntaxError : Error, Equatable, CustomStringConvertible {
   public let description:String
+  public var token: Token?
 
   public init(_ description:String) {
     self.description = description
   }
+  
+  public func contextAwareError(templateName: String?, templateContent: String) -> TemplateSyntaxError? {
```
kylef: I'd prefer to keep this logic out of the error class itself. I'd keep `TemplateSyntaxError` as a [POO](https://en.wikipedia.org/wiki/Plain_old_Java_object) without additional logic included.

I'd prefer to see the error containing all of the stack frames and contextual information and a separate renderer can be capable of rendering it in different ways.

There will be separate renders for ANSI, plain text, HTML, Bugsnag reporter, etc (eventually).

Sources/Lexer.swift
```diff
@@ -151,4 +162,21 @@ extension String {
     let last = findLastNot(character: character) ?? endIndex
     return self[first..<last]
   }
+  
+  func lineAndPosition(at range: Range<String.Index>) -> (content: String, number: Int, offset: String.IndexDistance) {
```
kylef: I'd move this over to the reporter too, and internally Stencil would use character(or possibly byte?)-level source maps everywhere.

kylef:
As an example I just mocked up how a HTML/Bugsnag reporter might look like for the errors. This might help demonstrate the need for multiple frames. In this case it would be possible to integrate the actual Swift frames too into the error leading up to the `TemplateSyntaxError`.

messages image 502882220

ilyapuchka:
@kylef yeah, I think providing more context would be really nice, but I guess it can be done additionally to associating error to specific token that failed to be parsed or node that failed to be rendered. I think that's the main pain point currently - it's hard to see where error comes from. Then we can add up context information to cover more complicated cases like one you described.

Regarding error reporter - I'll try to play with this idea 👍

ilyapuchka:
Something like this maybe:

```swift
protocol ErrorReporter: class { // concrete type to use to be set in Environment
  init(context: Context) // some context type with template name and content just for now
  func report(error: TemplateSyntaxError, for token: Token) throws // we might need the same method for some other structures like filters or nodes, or we could use some common protocol for them
  func contextAwareError(context: Context) -> TemplateSyntaxError?
}

class SimpleErrorReporter: ErrorReporter {

  func report(error: TemplateSyntaxError, for token: Token) throws {
    throw error.contextAwareError(context: context, token: token) ?? error
  }

  func contextAwareError(context: Context) -> TemplateSyntaxError? {
    ...
  }
}
```

ilyapuchka:
@kylef I've made more changes and now I think it looks more into direction of what you described in your previous comments.
This still does not cover errors that occur during rendering (only parsing is covered for now), and I'd like your input about stack trace output for implemented console error reporter.
Let me know what you think 😀

ilyapuchka:
@kylef will be be able to look at this again?

ilyapuchka:
@kylef I think this is ready now. I've also implemented capturing frames, i.e. for error from extended template reporter will print error like this:

```
invalid-base.html:2:24: error: Unknown filter 'unknown'
{% block body %}Body {{ target|unknown }} {% endblock %}
                        ^~~~~~~~~~~~~~

invalid-child-super.html:1:0: error: Unknown filter 'unknown'
{% extends "invalid-base.html" %}
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ilyapuchka:
I'm closing this as I accidentally removed my fork of the repo from which I created a pr, so now it does not allow to update it 😓

PR #149, opened 2017-11-15
Fix Swift 4 compilation error

dcwatson:
Installing 0.10.0 on Xcode 9.1/Swift 4 results in a compilation error to the effect of "Swift 4 subscripts returning string obsoleted". Not sure if this was a more recent change in 9.1, since I see you're already testing on Xcode 9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ matrix:
     - os: osx
       osx_image: xcode9
       env: SWIFT_VERSION=4.0
+    - os: osx
+      osx_image: xcode9.1
+      env: SWIFT_VERSION=4.0
     - os: linux
       env: SWIFT_VERSION=3.1.1
     - os: linux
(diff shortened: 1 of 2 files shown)


kylef:
@dcwatson, Thanks for the PR. Would you be able to add an extra testing matrix for Xcode 9.1 since this problem is only with that version of Xcode (if I understand correctly). You can copy the lines from https://github.com/kylef/Stencil/blob/master/.travis.yml#L6-L8 and update for Xcode 9.

OneSman7:
Waiting for this to be released in pod.
For now will make these chages directly in library code.

dcwatson:
Welp, the xcode9.1 tests ran successfully, not sure why the Swift 4 linux tests started failing, I didn't change anything else...

dcwatson:
Thanks! And thanks for the project, it's feels like home as a Django guy getting his feet wet with Swift.

PR #151, opened 2017-11-28
Add a protocol for keying into variables.

jimstudt:
On non-macOS platforms key value coding can not be used to extract values from variables. The introspection strategy works, but only for fields of the topmost class. Superclass fields are not visible. This makes any kind of hierarchical classes problematic.

This PR adds a protocol with a single function which works like the key value coding `func value(key)->Any?`. 

The advantages are:

 - cross platform safe
 - allows the Stencil templates to use names which are different from the internal implementation names of the underlying Swift files
 - allow the Swift types to limit what they export to the template engine
 - faster than KVC or introspection

I've placed the RenderKeyed protocol above KVC or it would not be used on macOS.

### DIFF
--- a/Sources/Node.swift
+++ b/Sources/Node.swift
@@ -56,6 +56,9 @@ public protocol Resolvable {
   func resolve(_ context: Context) throws -> Any?
 }
 
+public protocol RenderKeyed {
+  func value( forRenderKey key: String) -> Any?
+}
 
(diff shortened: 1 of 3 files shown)


ilyapuchka:
@jimstudt #152 should solve superclass properties issue without need for extra protocol

PR #153, opened 2017-11-29
fixed swift 4 warnings

ilyapuchka:


ilyapuchka:
Closing as this actually fails when using swift 3.1.1 but does not feel like checking swift version in every case worths it.

PR #160, opened 2017-12-22
Added method to register boolean filters

ilyapuchka:
This was touched on in the discussion here https://github.com/SwiftGen/StencilSwiftKit/pull/69 and we are extensively using such approach in Sourcery for our [custom filters](https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Generating/Template/Stencil/StencilTemplate.swift#L172). Would be nice if it would be available out of the box.

Benefit of such approach compared to using `not`, as mentioned in the discussion, is that it works with filters chaining.


Tests/StencilTests/FilterSpec.swift
```diff
@@ -182,5 +182,24 @@ func testFilter() {
...
+          return value > 0
+        }
+        return nil
+      }
+
+      let result = try Template(templateString: "{{ value|isPositive }}")
+        .render(Context(dictionary: ["value": 1], environment: Environment(extensions: [repeatExtension])))
+      try expect(result) == "true"
+      
+      let negativeResult = try Template(templateString: "{{ value|!isPositive }}")
```
kylef: I think `isNegative` would be far clearer name for the filter. I'd suggest that the API allows defining the boolean name instead of `!{name}` to promote clearer Stencil template filters.

ilyapuchka: This is just test example, of course `isNegative` makes more sense. What d you mean by `API allows defining the boolean name`?

kylef: The `registerBooleanFilter` API could allow supplying an opposite filter name. So for example if you wanted to make a boolean `isPositive` and `isNegative` this API would allow you to do so.

ilyapuchka: Agree, it will make it more generic and will still allow usage `!` for those tools which want it (i.e. Sourcery) 👍

ilyapuchka: @kylef done

yonaskolb: I agree allowing a custom negative name is nice, but perhaps it could be an optional parameter instead of forcing users to come up with an opposite name?

ilyapuchka: @yonaskolb I was thinking about that but decided not to enforce anything by default. One can write a simple helper for that in their implementation.

yonaskolb: A custom helper would need to know how to negate a boolean though, instead of having a standard defined by Stencil itself. If the negative name is nil we can provide that standard way

ilyapuchka: I mean helper like this:

```swift
extension Extension {
  func registerBooleanFilter(name: String, filter: @escaping (Any?) throws -> Bool?) {
    registerBooleanFilter(name: name, negativeName: "!\(name)", filter: filter)
  }
}
```

yonaskolb: Yeah, just thought if it was built in they wouldn't have to do that and decide which name and syntax they should use (`!`, `not`, `negative` etc). We could standardise it for them (in this case using `!`)

```swift
/// negativeFilterName defaults to !name
public func registerBooleanFilter(name: String, negativeFilterName: String? = nil, filter: @escaping (Any?) throws -> Bool?) {
    filters[name] = .simple(filter)
    filters[negativeFilterName ?? "!\(name)"] = .simple {
      guard let result = try filter($0) else { return nil }
      return !result
    }
   }
```

ilyapuchka: @djbe @AliSoftware any feedback on that?

AliSoftware: I think whenever we declare a boolean filter the !name should always exist. Providing another name for the negation should be possible but optional.

This means that if you register isPositive then !isPositive should always exist. You could optionally provide an alternate name for the opposite filter e.g. isNegative and in that case isPositive, isNegative, !isPositive, !isNegative should all exist

I'd go in that direction for two reasons:
1. Templates writers shouldn't bother that those `!` are actually part of the filter names. When they write `!isNegative` they think that `!` is an operator and `isNegative` is the filter name. So they'd expect `!` to work on _all_ boolean filters, not just the ones that don't have a custom inverse
- When in compound expressions, especially now that we'll support parentheses in expressions with recent PR, some might find more clear in certain context to explicitly use `! isPositive` rather than `isNegative`

Last but not least, isPositive shouldn't be the opposite of isNegative anyway. If we have two filters with those names in Stencil, I for one would expect `isPositive(0)` and `isNegative(0)` to both return the same. So I'd argue that this filter isn't even a good candidate for a custom name (at least not this one)

ilyapuchka: @AliSoftware if one registers filter in extension using this new method it will add a counterpart filter, so it's not up to template writer, but the author of extension to choose how filter is registered and how counterpart is named. I don't think Stencil should enforce using `!` in this context unless it becomes a "first-class" in boolean expression (and supplement `not` operator). Some tools may prefer using `not{filter}` instead and this API will allow them to do that.

ilyapuchka:
@kylef @yonaskolb @djbe can we move on with this please?

ilyapuchka:
@djbe updated docs, also removed `Boolean` from method name as it is unambiguous anyway

PR #163, opened 2017-12-23
For loop improvements

ilyapuchka:
This PR contains several improvements for `for` tag:

- fixed using spaces between loop variables. Before it was failing parsing.

```
{% for key, value in dict %}
```

- added subscripting arrays and dictionaries using loop variable for index or key:

```
{% for index, value in array %}
{{ index }}: {{ array.index }}
{% endfor %}
```

```
{% for key in dict %}
{{ key }}: {{ dict.key }}
{% endfor %}
```

- added support for iterating arrays with index and value. This (almost) removes need for `forloop.counter` and `forloop.counter0` variables.

```
{% for index, value in array %}
{{ index }}: {{ value }}
{% endfor %}
```
- added support for iterating tuples, similar to iterating dictionaries:

```
{% for label, value in tuple %}
{{ label }}: {{ value }}
{% endfor %}
```

```
{% for label in tuple %}
{{ label }}: {{ tuple.label }}
{% endfor %}
```

- add support for subscripting tuples with labels and indexes. I.e. tuple `(one: 1, two: 2)` can be subscripted with labels: `{{ tuple.one }} {{ tuple.two }}` or with indexes: `{{ tuple.0 }} {{ tuple.1 }} `

- added `break` and `continue` tags for breaking/continuing current loop (I started to implement labeled loops, but it will require a bit more effort).

```
{% for item in items %}
  {{ item }}
  {% if condition %}{% break %}{% endif %}
{% endfor %}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   - added support for iterating over tuple elements and subscripting it with label or index
   - added support for iterating array with index and value
   - fixed using spaces between loop variables
-  - added support for subscripting dictionary using loop variable for key
+  - added support for subscripting arrays and dictionaries using loop variables for index or key
   - added `break` and `continue` tags to break or continue current loop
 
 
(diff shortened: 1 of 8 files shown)


ilyapuchka:
@kylef I'll break this into smaller separate prs so that it's easier for you to review, I expected it to be smaller, but it grew fast =))

PR #164, opened 2017-12-23
Resolvable boolean expressions

ilyapuchka:
This PR adds ability to resolve boolean expressions the same way as variables like `{{ this == that }}`

This have a potential to simplify templates when there is a need to evaluate boolean expression and output its result, true or false, in a template, i.e. when generating html to enable or disable html elements. Instead of  `{% if this == that %}true{% else %}false{% endif %}` it will be possible to write just `{{ this == that }}`

Also this PR adds support for parsing static boolean expressions, i.e. if `{% if true %}`. There was implementation for such expression and tests for it, but it was never actually parsed. It may not be very important, but can be helpful during template development.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,12 @@
   [#215](https://github.com/stencilproject/Stencil/pull/215)
 
 - Adds support for using spaces in filter expression
-  [Ilya Puchka](https://github.com/yonaskolb)
+  [Ilya Puchka](https://github.com/ilyapuchka)
   [#178](https://github.com/stencilproject/Stencil/pull/178)
+
+- Now boolean expressions results can be rendered, i.e `{{ name == "John" }}` will render `true` or `false` depending on the evaluation result 
+  [Ilya Puchka](https://github.com/ilyapuchka)
+  [#164](https://github.com/stencilproject/Stencil/pull/164)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,10 @@
   object `item = ["name": "John"]`, then `{{ item[key] }}` will evaluate to "John".  
   [David Jennes](https://github.com/djbe)
   [#215](https://github.com/stencilproject/Stencil/pull/215)
-
 - Adds support for using spaces in filter expression
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#178](https://github.com/stencilproject/Stencil/pull/178)
-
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -13,9 +13,13 @@
   [#215](https://github.com/stencilproject/Stencil/pull/215)
 
 - Adds support for using spaces in filter expression
-  [Ilya Puchka](https://github.com/yonaskolb)
+  [Ilya Puchka](https://github.com/ilyapuchka)
   [#178](https://github.com/stencilproject/Stencil/pull/178)
 
+- Now boolean expressions results can be rendered, i.e `{{ name == "John" }}` will render `true` or `false` depending on the evaluation result
```
djbe: Forgot the `.` and 2 spaces. Also remove the empty line above.

Sources/Expression.swift
```diff
@@ -40,9 +47,13 @@ final class VariableExpression: Expression, CustomStringConvertible {
...
   }
-
+  
+  func resolve(_ context: Context) throws -> Any? {
+    return try variable.resolve(context)
+  }
+  
   /// Resolves a variable in the given context as boolean
-  func resolve(context: Context, variable: Resolvable) throws -> Bool {
+  func evaluate(context: Context) throws -> Bool {
```
djbe: Weird, I think I've seen this change a few times in other PRs? 😕😆

djbe:

Tests/StencilTests/ExpressionSpec.swift
```diff
@@ -136,163 +147,172 @@ func testExpressions() {
         }
       }
 
+      func expectExpression(with components: [String], context: [String: Any], toBe expected: Bool) throws {
+        let expression = try parseExpression(components: components, tokenParser: parser)
+        try expect(expression.evaluate(context: Context(dictionary: context))) == expected
```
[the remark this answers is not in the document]
djbe: Good idea!

Sources/Expression.swift
ilyapuchka: maybe, some features required similar changes

ilyapuchka:
@djbe @kylef comments addressed, can you look again?

djbe:
@ilyapuchka why close this? The docs are there, it should be GTM.

ilyapuchka:
there were conflicts and tests started to fail for some reason after I merged, I'll need to to investigate that but I also don't feel like this is a top priority considering other open PRs

djbe:
Which tests? CI succeeds for the latest commit in this PR.

ilyapuchka:
Only because I didn't push 😉

PR #165, opened 2017-12-25
Added support for brackets in boolean expressions

ilyapuchka:
Currently there is no way to alter operations precedence in boolean expression, so it can force users to use cascades of ifs. With this it should become simpler.

Expression in brackets is parsed recursively as special case of `IfToken` that represents sub-expression which is evaluated similar to variable token.


Sources/IfTag.swift
```diff
@@ -100,21 +107,74 @@ final class IfExpressionParser {
...
+
+  static func validateBracketsBalance(components: [String]) throws {
+    guard try components.reduce(0, {
+      var bracketsBalance = $0
+      if $1 == "(" { bracketsBalance += 1 }
+      else if $1 == ")" { bracketsBalance -= 1 }
+      if bracketsBalance < 0 { throw TemplateSyntaxError("unbalanced brackets") }
+      return bracketsBalance
+    }) == 0 else {
+      throw TemplateSyntaxError("unbalanced brackets")
```
AliSoftware: Should we give a more explicit error, like "unbalanced brackets: missing closing bracket" here and "unbalanced brackets: too many closing brackets" above when <0?

Sources/IfTag.swift
```diff
@@ -100,21 +107,74 @@ final class IfExpressionParser {
...
+        return .variable(try tokenParser.compileFilter(component))
+      } else {
+        return nil
       }
-
-      return .variable(try tokenParser.compileFilter(component))
     }
   }
 
+  static func subExpression(from index: Int, components: [String], tokenParser: TokenParser) throws -> (Expression, [String]) {
```
AliSoftware: Wouldn't it be better performance and memory-wise to use an ArraySlice here for components, to prevent rebuilding a new sub-array from the full array on every recursion level? After all, that's exactly what ArraySlices are designed for 😉

That would allow you to get rid of 'Array(…)` wrapping 2 lines below too.
Of course you'd need to replace your `components.enumerated()` on line 133 by `zip(components, components.indices)` instead, which is generally better anyway (see Soroush's article about enumerated) because not all startIndices are 0.

ilyapuchka: Can't say anything about performance, but the only thing it will really change is move `suffix` to caller site, because `IfExpressionParser` accepts `Array`. But it allows to remove one function parameter, so I'll make the change still.

Speaking of enumeration I don't see what `zip` will improve in this particular case. I find `enumerated` use to be more clear here.

AliSoftware: It's not about readability but about correctness (see Soroush's article), especially if we move to use slices (because they're the type to use when you recurse on subexpression, as they avoid a ton of memory reallocation, and that's exactly what we're doing here so that's a textbook use case) and enumerated will lead to wrong indexes with slices unlike zip (as explained by Soroush)

ilyapuchka: Ok, if this type only accepts `ArraySlice` in constructor it makes sense 👍

Sources/IfTag.swift
```diff
@@ -100,21 +107,72 @@ final class IfExpressionParser {
...
+    try IfExpressionParser.validateBracketsBalance(components: components)
+    return try IfExpressionParser(components: ArraySlice(components), tokenParser: tokenParser)
+  }
+
+  static func validateBracketsBalance(components: [String]) throws {
+    guard try components.reduce(0, {
+      var bracketsBalance = $0
+      if $1 == "(" { bracketsBalance += 1 }
+      else if $1 == ")" { bracketsBalance -= 1 }
+      if bracketsBalance < 0 { throw TemplateSyntaxError("unbalanced brackets: too many closing brackets") }
```
AliSoftware: Sorry to be nitpicking about code style, but if we soon add SwiftLint to harmonize the code style of all contributors it will raise a violation here probably: even for one-liners like that it'll suggest to go to the next line after opening brace and before closing brace. Besides just being strict with style (which I agree can be annoying sometimes) this has the advantage of allowing to debug step by step, entering inside the scope of needed during debug.

Sources/IfTag.swift
```diff
@@ -100,21 +107,72 @@ final class IfExpressionParser {
...
+          }
         }
+        return .variable(try tokenParser.compileFilter(component))
       }
-
-      return .variable(try tokenParser.compileFilter(component))
     }
   }
 
+  static func subExpression(from components: ArraySlice<String>, tokenParser: TokenParser) throws -> (Expression, Int) {
```
AliSoftware: Any reason why this isn't `private`?

Sources/IfTag.swift
```diff
@@ -100,21 +107,72 @@ final class IfExpressionParser {
...
+    var bracketsBalance = 1
+    let subComponents = components
+      .prefix(while: {
+        if $0 == "(" { bracketsBalance += 1 }
+        else if $0 == ")" { bracketsBalance -= 1 }
+        return bracketsBalance != 0
+      })
+
```
AliSoftware: Not really sure it would be worth it, but what about a small `class Balancer` helper for every time we have to balance parentheses and/or brackets? Could serve here, in the coffee above too, and in the other PR from @djbe about indirect reference…

```swift
class Balancer {
  var balance = 0
  let opening: Character
  let closing: Character
  init(pair: (Character, Character), initialBalance: Int = 0) {
    opening = pair.0
    closing = pair.1
    balance = initialBalance
  }
  // convenience constructors
  static func brackets(initial: Int = 0) -> Balancer {
    return Balancer(pair: ("[", "]"), initialBalance: initial)
  }
  // And same for parentheses

  func parse(_ char: Character) throws -> Int {
    // … parse the char, increment/decrement the balance if it matches the opening/closing chars, throws if balance < 0, return new balance
  }
}
```
For example here you could then create a balancer then use `components.prefix(while: { balancer.parse($0) != 0)`

_(all that code typed from my phone so no guarantee of being correct or typo-free)_

I'll let you discuss if it's worth adding it? but could be nice for various places where we do that kind of logic, in this PR and others

Sources/IfTag.swift
```diff
@@ -100,21 +107,72 @@ final class IfExpressionParser {
...
+      if $1 == "(" { bracketsBalance += 1 }
+      else if $1 == ")" { bracketsBalance -= 1 }
+      if bracketsBalance < 0 { throw TemplateSyntaxError("unbalanced brackets: too many closing brackets") }
+      return bracketsBalance
+    }) == 0 else {
+      throw TemplateSyntaxError("unbalanced brackets: missing closing bracket")
+    }
+  }
+  
+  private init(components: ArraySlice<String>, tokenParser: TokenParser) throws {
```
AliSoftware: 👌

Sources/IfTag.swift
ilyapuchka: Good idea but I don't think it will be useful in #215 as validation there is done during parsing, not as a separate step. We can do the same here and remove validation step.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Added support for resolving superclass properties for not-NSObject subclasses
 - The `{% for %}` tag can now iterate over tuples, structures and classes via
   their stored properties.
+- You can now use brackets in boolean expressions to change operators precedence
```
djbe: Could you update the entry format to match the rest? (`.`, 2 spaces, credit yourself and ref. this PR)

Sources/IfTag.swift
```diff
@@ -100,21 +107,68 @@ final class IfExpressionParser {
...
+      if component == "(" {
+        bracketsBalance += 1
+        let (expression, parsedCount) = try IfExpressionParser.subExpression(
+          from: components.suffix(from: index + 1),
+          tokenParser: tokenParser
+        )
+        parsedComponents.formUnion(Set(index...(index + parsedCount)))
+        return .subExpression(expression)
+      }
+      else if component == ")" {
```
djbe: Why isn't this on the same line as the `}`?

Sources/IfTag.swift
```diff
@@ -100,21 +107,68 @@ final class IfExpressionParser {
...
-
-      return .variable(try tokenParser.compileFilter(component))
     }
   }
 
+  private static func subExpression(from components: ArraySlice<String>, tokenParser: TokenParser) throws -> (Expression, Int) {
+    var bracketsBalance = 1
+    let subComponents = components
+      .prefix(while: {
+        if $0 == "(" { bracketsBalance += 1 }
```
djbe: Not a big fan of having the content of the `if` on the same line, and pretty sure `swiftlint`'ll complain about this too.

Sources/IfTag.swift
```diff
@@ -100,21 +107,68 @@ final class IfExpressionParser {
...
+        if bracketsBalance < 0 {
+          throw TemplateSyntaxError("'if' expression error: missing opening bracket")
+        }
+        parsedComponents.insert(index)
+        return nil
+      } else {
+        parsedComponents.insert(index)
+        if let op = findOperator(name: component) {
+          switch op {
+          case .infix(let name, let bindingPower, let cls):
```
djbe: I had to look up what the `cls` value was in the `Operator` enum. Maybe `..., let operator` woud be clearer?

Tests/StencilTests/ExpressionSpec.swift
```diff
@@ -295,6 +295,42 @@ func testExpressions() {
...
+            .toThrow(TemplateSyntaxError("'if' expression error: missing closing bracket"))
+          try expect(parseExpression(components: ["(", "lhs", "and", "rhs", ")", ")"], tokenParser: parser))
+            .toThrow(TemplateSyntaxError("'if' expression error: missing opening bracket"))
+          try expect(parseExpression(components: ["(", "lhs", "and", ")"], tokenParser: parser))
+            .toThrow(TemplateSyntaxError("'if' expression error: end"))
+          try expect(parseExpression(components: ["(", "and", "rhs", ")"], tokenParser: parser))
+            .toThrow(TemplateSyntaxError("'if' expression error: infix operator 'and' doesn't have a left hand side"))
+        }
+      }
+
```
djbe: unneeded empty new line

djbe:

ilyapuchka:
@djbe comments addressed, can you look again?

docs/builtins.rst
```diff
@@ -149,6 +149,19 @@ Will be treated as:
 
     one or (two and three)
 
+You can use brackets to change operators precedence. For example:
```
djbe: You can use parentheses to change operator precedence. For example:

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 - Now requires Swift 4.1 or newer.  
   [Yonas Kolb](https://github.com/yonaskolb)
   [#228](https://github.com/stencilproject/Stencil/pull/228)
+- You can now use brackets in boolean expressions to change operators precedence.
```
djbe: Weird that I didn't catch this before. Generally "brackets" (in en-US) is used to refer to "[ ]", whereas "parentheses" is used for "( )". Also pretty sure that in this situation it's "operator" (without an `s`).

> You can now use parentheses in boolean expressions to change operator precedence.

ilyapuchka:
@djbe fixed docs, good catch with brackets vs parentheses 👍

djbe:
Does this still need a rebase? (github says it needs an update)

ilyapuchka:
@djbe I'll update and sqaush

PR #166, opened 2017-12-26
Flaky tests fixed

ilyapuchka:


Sources/Context.swift
```diff
@@ -4,7 +4,7 @@ public class Context {
 
   public let environment: Environment
 
-  init(dictionary: [String: Any]? = nil, environment: Environment? = nil) {
+  init(dictionary: [String: Any?]? = nil, environment: Environment? = nil) {
```
kylef: Wouldn't `Optional` be of type `Any` already? I am not sure that the `Any?` adds, and what it is trying to fix. The commit message mentions it is fixing some compiler warnings, which version of the Swift compiler are you seeing the warnings? I'm not seeing any here from 4.0.1 locally.

ilyapuchka: One of the cases was that as context dictionaries are defined with `Any?` I was using `[String: Any?]` type to push to the context too. Then compiler was giving a warning "Expression implicitly coerced from Any? to Any". Another case is that having Any? here will let to use `nil` values in dictionaries passed in these methods which may simplify constructing of such dictionaries by avoiding unwrapping optionals. And I think it's easier with matching type annotations.

ilyapuchka:
@kylef as you can see I base all other prs on this pr to avoid tests failures. I hate it though, so as soon as you will merge it I'll rebase other prs =)

PR #167, opened 2017-12-26
Errors logs improvements

ilyapuchka:
Original PR #146 

Example error logs:

```
1:9: error: 'for' statements should use the following syntax 'for x in y where condition'.
Hello {% for name in %}{{ name }}, {% endfor %}!
         ^~~~~~~~~~~
```

```
invalid-base.html2:24: error: filter error
{% block body %}Body {{ target|unknown }} {% endblock %}
                        ^~~~~~~~~~~~~~

invalid-child-super.html2:25: error: filter error
{% block body %}Child {{ block.super }}{% endblock %}
                         ^~~~~~~~~~~
```

### DIFF
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -3,6 +3,10 @@ public class Context {
   var dictionaries: [[String: Any?]]
 
   public let environment: Environment
+  
+  public var errorReporter: ErrorReporter {
+    return environment.errorReporter
+  }
 
   init(dictionary: [String: Any]? = nil, environment: Environment? = nil) {
     if let dictionary = dictionary {
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -4,10 +4,6 @@ public class Context {
 
   public let environment: Environment
   
-  public var errorReporter: ErrorReporter {
-    return environment.errorReporter
-  }
-
   init(dictionary: [String: Any]? = nil, environment: Environment? = nil) {
     if let dictionary = dictionary {
       dictionaries = [dictionary]
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -9,7 +9,7 @@ public struct Environment {
               extensions: [Extension]? = nil,
               templateClass: Template.Type = Template.self,
               errorReporter: ErrorReporter = SimpleErrorReporter()) {
-    
+
     self.templateClass = templateClass
     self.errorReporter = errorReporter
     self.loader = loader
@@ -41,11 +41,11 @@ public struct Environment {
     let template = templateClass.init(templateString: string, environment: self)
     return try render(template: template, context: context)
   }
-  
+
   func render(template: Template, context: [String: Any]?) throws -> String {
     // update template environment as it can be created from string literal with default environment
     template.environment = self
     return try template.render(context)
   }
-  
+
 }
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -3,9 +3,15 @@ public struct Environment {
   public var extensions: [Extension]
 
   public var loader: Loader?
+  public var errorReporter: ErrorReporter
 
-  public init(loader: Loader? = nil, extensions: [Extension]? = nil, templateClass: Template.Type = Template.self) {
+  public init(loader: Loader? = nil,
+              extensions: [Extension]? = nil,
+              templateClass: Template.Type = Template.self,
+              errorReporter: ErrorReporter = SimpleErrorReporter()) {
+    
     self.templateClass = templateClass
+    self.errorReporter = errorReporter
     self.loader = loader
     self.extensions = (extensions ?? []) + [DefaultExtension()]
   }
@@ -28,11 +34,21 @@ public struct Environment {
 
   public func renderTemplate(name: String, context: [String: Any]? = nil) throws -> String {
     let template = try loadTemplate(name: name)
-    return try template.render(context)
+    return try render(template: template, context: context)
   }
 
   public func renderTemplate(string: String, context: [String: Any]? = nil) throws -> String {
     let template = templateClass.init(templateString: string, environment: self)
-    return try template.render(context)
+    return try render(template: template, context: context)
+  }
+  
+  func render(template: Template, context: [String: Any]?) throws -> String {
+    errorReporter.context = ErrorReporterContext(template: template)
+    do {
+      return try template.render(context)
+    } catch {
+      try errorReporter.report(error: error)
+      return ""
+    }
   }
 }
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -51,4 +51,9 @@ public struct Environment {
       return ""
     }
   }
+  
+  var template: Template? {
+    return errorReporter.context?.template
+  }
+  
 }
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -3,15 +3,12 @@ public struct Environment {
   public var extensions: [Extension]
 
   public var loader: Loader?
-  public var errorReporter: ErrorReporter
 
   public init(loader: Loader? = nil,
               extensions: [Extension]? = nil,
-              templateClass: Template.Type = Template.self,
-              errorReporter: ErrorReporter = SimpleErrorReporter()) {
+              templateClass: Template.Type = Template.self) {
 
     self.templateClass = templateClass
-    self.errorReporter = errorReporter
     self.loader = loader
     self.extensions = (extensions ?? []) + [DefaultExtension()]
   }
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -43,7 +43,7 @@ public struct Environment {
   }
   
   func render(template: Template, context: [String: Any]?) throws -> String {
-    // update temaplte environment as it cen be created from string literal with default environment
+    // update template environment as it can be created from string literal with default environment
     template.environment = self
     return try template.render(context)
   }
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -45,30 +45,7 @@ public struct Environment {
   func render(template: Template, context: [String: Any]?) throws -> String {
     // update temaplte environment as it cen be created from string literal with default environment
     template.environment = self
-    errorReporter.context = ErrorReporterContext(template: template)
-    do {
-      return try template.render(context)
-    } catch {
-      throw errorReporter.reportError(error)
-    }
-  }
-  
-  var template: Template? {
-    return errorReporter.context?.template
-  }
-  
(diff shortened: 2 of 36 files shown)


Sources/Context.swift
```diff
@@ -3,8 +3,12 @@ public class Context {
   var dictionaries: [[String: Any?]]
 
   public let environment: Environment
+  
+  public var errorReporter: ErrorReporter {
+    return environment.errorReporter
+  }
 
-  init(dictionary: [String: Any]? = nil, environment: Environment? = nil) {
+  init(dictionary: [String: Any?]? = nil, environment: Environment? = nil) {
```
kylef: I've seen this change in multiple pull requests. Wanna split this out into a separate PR for discussion?

ilyapuchka: it's in #166 , I just based all other branches on its branch so that tests don't randomly fail, that's why you see them in each pr =)

ilyapuchka:
@kylef I've simplified this a lot by storing filenames and token positions at parsing time instead of calculating them at rendering time, now it looks much closer to what you've started in #82 .

ilyapuchka:
@kylef this started to fail on swift 3.1 with some creepy  compiler error. Maybe we should just drop it already?

DivineDominion:
Whoa this looks cool. I'd like to +1 here so somebody from the team takes a look :)

djbe:
@ilyapuchka Any idea how hard it'd be to rebase/rework this on master for you?
While this may not seem a "flashy" issue with fancy new features, it could end up being the most useful PR of all those that are open right now for template writers.

ilyapuchka:
@djbe I've updated this with the latest master. If you are going to merge any PRs please merge this first.

Tests/StencilTests/ForNodeSpec.swift
```diff
@@ -98,15 +98,15 @@ func testForNode() {
 
     $0.it("renders the given nodes while filtering items using where expression") {
         let nodes: [NodeType] = [VariableNode(variable: "item"), VariableNode(variable: "forloop.counter")]
-        let `where` = try parseExpression(components: ["item", ">", "1"], tokenParser: TokenParser(tokens: [], environment: Environment()))
+      let `where` = try parseExpression(components: ["item", ">", "1"], tokenParser: TokenParser(tokens: [], environment: Environment()), token: .text(value: "", at: .unknown))
```
djbe: Indentation?

Tests/StencilTests/ForNodeSpec.swift
```diff
@@ -98,15 +98,15 @@ func testForNode() {
...
         let node = ForNode(resolvable: Variable("items"), loopVariables: ["item"], nodes: nodes, emptyNodes: [], where: `where`)
         try expect(try node.render(context)) == "2132"
     }
 
     $0.it("renders the given empty nodes when all items filtered out with where expression") {
         let nodes: [NodeType] = [VariableNode(variable: "item")]
         let emptyNodes: [NodeType] = [TextNode(text: "empty")]
-        let `where` = try parseExpression(components: ["item", "==", "0"], tokenParser: TokenParser(tokens: [], environment: Environment()))
+      let `where` = try parseExpression(components: ["item", "==", "0"], tokenParser: TokenParser(tokens: [], environment: Environment()), token: .text(value: "", at: .unknown))
```
djbe: Indentation?

Sources/ForTag.swift
```diff
@@ -6,54 +6,56 @@ class ForNode : NodeType {
...
     }
 
-    let resolvable = try parser.compileResolvable(components[3])
-
-    let `where` = hasToken("where", at: 4)
-      ? try parseExpression(components: Array(components.suffix(from: 5)), tokenParser: parser)
-      : nil
-
-    return ForNode(resolvable: resolvable, loopVariables: loopVariables, nodes: forNodes, emptyNodes:emptyNodes, where: `where`)
+    return ForNode(resolvable: resolvable, loopVariables: loopVariables, nodes: forNodes, emptyNodes:emptyNodes, where: `where`, token: token)
```
djbe: Space after the `:`.

Sources/ForTag.swift
```diff
@@ -6,54 +6,56 @@ class ForNode : NodeType {
...
   }
 
-  init(resolvable: Resolvable, loopVariables: [String], nodes:[NodeType], emptyNodes:[NodeType], where: Expression? = nil) {
+  init(resolvable: Resolvable, loopVariables: [String], nodes:[NodeType], emptyNodes:[NodeType], where: Expression? = nil, token: Token? = nil) {
```
djbe: Space after the `:`.

Sources/Lexer.swift
```diff
@@ -156,4 +179,24 @@ extension String {
...
+    var lineNumber: UInt = 0
+    var offset: Int = 0
+    var lineContent = ""
+
+    for line in components(separatedBy: CharacterSet.newlines) {
+      lineNumber += 1
+      lineContent = line
+      if let rangeOfLine = self.range(of: line), rangeOfLine.contains(range.lowerBound) {
+        offset = distance(from: rangeOfLine.lowerBound, to:
+          range.lowerBound)
```
djbe: bizarre spacing here?

Sources/Node.swift
```diff
@@ -1,35 +1,31 @@
...
+
+  /// Reference to this node's token
+  var token: Token? { get }
 }
 
 
 /// Render the collection of nodes in the given context
 public func renderNodes(_ nodes:[NodeType], _ context:Context) throws -> String {
-  return try nodes.map { try $0.render(context) }.joined(separator: "")
+  return try nodes.map({
```
djbe: Those `()` aren't needed for the `map`.

Sources/Parser.swift
```diff
@@ -110,15 +116,42 @@ public class TokenParser {
     return filtersWithDistance.filter({ $0.distance == minDistance }).map({ $0.filterName })
   }
 
+  public func compileFilter(_ filterToken: String, containedIn containingToken: Token) throws -> Resolvable {
+    do {
+      return try FilterExpression(token: filterToken, parser: self)
+    } catch {
+      if var error = error as? TemplateSyntaxError, error.token == nil {
```
djbe: Might be good to use a guard here? It avoids 1 level of indentation.

Tests/StencilTests/EnvironmentSpec.swift
```diff
@@ -32,9 +39,304 @@ func testEnvironment() {
 
       try expect(result) == "here"
     }
+
+    func expectedSyntaxError(token: String, template: Template, description: String) -> TemplateSyntaxError {
+      let range = template.templateString.range(of: token)!
```
djbe: Always avoid implicit unwraps. You can replace this with a guard that falls back to an error with less information if the token isn't found. Or a `fatalError` (as this is a test, might as well break early).

Tests/StencilTests/EnvironmentSpec.swift
```diff
@@ -32,9 +39,304 @@ func testEnvironment() {
...
+      let sourceMap = SourceMap(filename: template.name, line: rangeLine)
+      let token = Token.block(value: token, at: sourceMap)
+      return TemplateSyntaxError(reason: description, token: token, stackTrace: [])
+    }
+
+    func expectError(reason: String, token: String,
+                     file: String = #file, line: Int = #line, function: String = #function) throws {
+      let expectedError = expectedSyntaxError(token: token, template: template, description: reason)
+      
+      let error = try expect(environment.render(template: template, context: ["names": ["Bob", "Alice"], "name": "Bob"]),
```
djbe: Both the declaration of this `func` as well as this line have some bizarre indentation. This also applies to the similar functions below.

djbe: This function is repeated quite a few times. That could be rewritten?

Tests/StencilTests/LexerSpec.swift
```diff
@@ -66,25 +69,26 @@ func testLexer() {
...
-        try expect(tokens[2]) == Token.variable(value: "name")
-        try expect(tokens[3]) == Token.block(value: "endif")
-        try expect(tokens[4]) == Token.text(value: ".")
+          "    if name\n" +
+          "     and\n" +
+          "    name\n" +
+          "%}{{\n" +
+          "name\n" +
+          "}}{%\n" +
+        "endif %}."
```
djbe: Indentation for this last line.

djbe:
Seeing as you'd rather merge this first, I'll review this one before the others.

Tests/StencilTests/EnvironmentSpec.swift
ilyapuchka: I'd prefer this to 6 arguments stacked on each other
this can be cleared up sometime later, this PR is big without it

Tests/StencilTests/EnvironmentSpec.swift
ilyapuchka: I would prefer failing here as it is crystal clear what is the reason than pollute the code with boilerplate guards especially considering that this is a test and we don't need graceful error handling or whatever.

ilyapuchka:
@djbe fixed indentations

Sources/Environment.swift
```diff
@@ -3,9 +3,15 @@ public struct Environment {
   public var extensions: [Extension]
 
   public var loader: Loader?
+  public var errorReporter: ErrorReporter
+
+  public init(loader: Loader? = nil,
+              extensions: [Extension]? = nil,
+              templateClass: Template.Type = Template.self,
+              errorReporter: ErrorReporter = SimpleErrorReporter()) {
```
kylef: If I understand correctly, the errorReporter property is only used by context's errorReporter property. But that isn't used anywhere (I may have missed this)?

What is the reason that the Environment and Context should hold the errorReporter? If I understand how this is intended to be used, a consumer of the library would catch an error and pass it to the errorReporter themselves? In that case I think it would make sense to remove the property from Environment and Context leaving the consumer free to create their own ErrorReporter as desired. Decoupling these a little bit.

ilyapuchka:
@kylef this is probably a leftover from when I was using it to render error in inheritance nodes. Not needed now, removed in 92ebfe5

Tests/StencilTests/EnvironmentSpec.swift
djbe: A `guard` with a `fatalError` and a message then. It's not because it's a test that it shouldn't be good code.

ilyapuchka: I think this code is good enough, but I will do that if that's what it takes to merge this.

ilyapuchka: @djbe done in 96a004e

ilyapuchka:
@kylef you'd like to add something?

djbe:
Seeing as there's no more feedback, and this PR is GTM for me, we can go on to the next PRs 🎉

PR #169, opened 2017-12-27
Fixed using spaces between loop variables

ilyapuchka:
Though docs [say](https://stencil.fuller.li/en/latest/builtins.html#built-in-tags) that you can do `for key, value in dict` in fact it was failing parsing.


ilyapuchka:
Solved ins #178 in a more generic way

PR #170, opened 2017-12-27
Iterating arrays with their indexes and values

ilyapuchka:
This allows to enumerate arrays by their indexes and values, not just by their values, instead of using `forloop` context variable.


kylef:
Rebased this on master, this is breaking the behaviour of iterating over tuples. We will have to rethink how to implement this one.

ilyapuchka:
@kylef as we have counters now I think we can close this

ilyapuchka:
Probably we should introduce another way to unpack tuples, not sure if using for loop for that is good. Then we will be able to do this feature. But it will be breaking change.

PR #175, opened 2017-12-27
Break, continue and loops' labels

ilyapuchka:
This PR adds `break` and `continue` tags for loops. They can be labeled, like in Swift, or not.

```
{% for item in items %}
  {{ item }}{% if condition %}{% break %}{% endif %}
{% endfor %}
```

Labeled loops:

```
{% outer: for item in items %}
  {% for key in keys %}
    {{ item.key }}{% if condition %}{% break outer %}{% endif %}
  {% endfor %}
{% endfor %}
```

Also with labels it's now possible to access outer loop context from inner loop:

```
{% outer: for item in items %}
  {% for key in keys %}
   {{ forloop.counter }}-{{ forloop.outer.counter }}
  {% endfor %}
{% endfor %}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#158](https://github.com/stencilproject/Stencil/issues/158)
   [#182](https://github.com/stencilproject/Stencil/pull/182)
+- Added `break` and `continue` tags to break or continue current loop.  
+  [Ilya Puchka](https://github.com/ilyapuchka)
+  [#175](https://github.com/stencilproject/Stencil/pull/175)
+- You can now access outer loop's scope by labeling it: `{% outer: for ... %}... {% for ... %} {{ outer.counter }} {% endfor %}{% endfor %}`.  
+  [Ilya Puchka](https://github.com/ilyapuchka)
+  [#175](https://github.com/stencilproject/Stencil/pull/175)
 
 ### Deprecations
 
--- a/Sources/Stencil/Extension.swift
+++ b/Sources/Stencil/Extension.swift
@@ -56,6 +56,8 @@ class DefaultExtension: Extension {
 
   fileprivate func registerDefaultTags() {
     registerTag("for", parser: ForNode.parse)
+    registerTag("break", parser: LoopTerminationNode.parse)
+    registerTag("continue", parser: LoopTerminationNode.parse)
     registerTag("if", parser: IfNode.parse)
     registerTag("ifnot", parser: IfNode.parse_ifnot)
     #if !os(Linux)
--- a/Sources/Stencil/ForTag.swift
+++ b/Sources/Stencil/ForTag.swift
@@ -6,10 +6,16 @@ class ForNode: NodeType {
   let nodes: [NodeType]
   let emptyNodes: [NodeType]
   let `where`: Expression?
+  let label: String?
   let token: Token?
 
   class func parse(_ parser: TokenParser, token: Token) throws -> NodeType {
-    let components = token.components
+    var components = token.components
+
+    var label: String?
+    if components.first?.hasSuffix(":") == true {
+      label = String(components.removeFirst().dropLast())
+    }
 
     func hasToken(_ token: String, at index: Int) -> Bool {
       components.count > (index + 1) && components[index] == token
@@ -52,6 +58,7 @@ class ForNode: NodeType {
       nodes: forNodes,
       emptyNodes: emptyNodes,
       where: `where`,
+      label: label,
       token: token
     )
   }
@@ -62,6 +69,7 @@ class ForNode: NodeType {
     nodes: [NodeType],
     emptyNodes: [NodeType],
     where: Expression? = nil,
+    label: String? = nil,
     token: Token? = nil
   ) {
     self.resolvable = resolvable
@@ -69,6 +77,7 @@ class ForNode: NodeType {
     self.nodes = nodes
     self.emptyNodes = emptyNodes
     self.where = `where`
+    self.label = label
     self.token = token
   }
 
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Added support for resolving superclass properties for not-NSObject subclasses
 - The `{% for %}` tag can now iterate over tuples, structures and classes via
   their stored properties.
+- Added `break` and `continue` tags to break or continue current loop
```
djbe: Please conform these to the formatting for other changelog entries (`.`, 2 spaces, credit yourself and link to this PR).

Sources/ForTag.swift
```diff
@@ -6,9 +6,15 @@ class ForNode : NodeType {
...
   let emptyNodes: [NodeType]
   let `where`: Expression?
+  let label: String?
 
   class func parse(_ parser:TokenParser, token:Token) throws -> NodeType {
-    let components = token.components()
+    var components = token.components()
+
+    var label: String? = nil
+    if components.first?.hasSuffix(":") == true {
```
djbe: No need for `== true`

Sources/ForTag.swift
```diff
@@ -124,25 +131,115 @@ class ForNode : NodeType {
...
-        let forContext: [String: Any] = [
+      for (index, item) in values.enumerated() {
+        var forContext: [String: Any] = [
           "first": index == 0,
           "last": index == (count - 1),
           "counter": index + 1,
-          "counter0": index,
-        ]
+          "counter0": index
+          ]
```
djbe: alignment?

Sources/ForTag.swift
```diff
@@ -124,25 +131,115 @@ class ForNode : NodeType {
 
     if !values.isEmpty {
       let count = values.count
+      var result = ""
+
+      var currentLoopContext = context["forloop"] as? [String: Any]
 
-      return try values.enumerated().map { index, item in
-        let forContext: [String: Any] = [
+      for (index, item) in values.enumerated() {
```
djbe: Replace the `enumerated()` with `zip(0..., values)`.

Sources/ForTag.swift
```diff
@@ -124,25 +131,115 @@ class ForNode : NodeType {
 
     if !values.isEmpty {
       let count = values.count
+      var result = ""
+
+      var currentLoopContext = context["forloop"] as? [String: Any]
```
djbe: I was a bit confused on how you were getting the previously existing loop contexts, maybe give this a more descriptive name? For example `existingLoopContext` or `parentLoopContext`?

Sources/ForTag.swift
```diff
@@ -124,25 +131,115 @@ class ForNode : NodeType {
...
           "counter": index + 1,
-          "counter0": index,
-        ]
+          "counter0": index
+          ]
+        if let label = label {
+          forContext[label] = forContext
+          forContext["label"] = label
+        }
+        if let label = currentLoopContext?["label"] as? String,
```
djbe: If I understand this correctly, this will only work for 1 parent labeled loop. If I have more than 1 parent labeled loop, I won't be able to access their context.

I'm not saying we should support such a use case (or not), but if we don't, we should document it clearly.

Sources/ForTag.swift
```diff
@@ -124,25 +131,115 @@ class ForNode : NodeType {
...
+        }
 
-        return try context.push(dictionary: ["forloop": forContext]) {
-          return try push(value: item, context: context) {
+        var shouldBreak: Bool = false
+        result += try context.push(dictionary: ["forloop": forContext]) {
+          let result = try push(value: item, context: context) {
             try renderNodes(nodes, context)
           }
+          shouldBreak = context[LoopTerminationNode.breakContextKey] != nil
```
djbe: Maybe move the shouldBreak calculation logic out to a small method to keep this part a bit cleaner.

Sources/ForTag.swift
```diff
@@ -124,25 +131,115 @@ class ForNode : NodeType {
...
     }
 
     return try context.push {
       try renderNodes(emptyNodes, context)
     }
   }
 }
+
+struct LoopTerminationNode: NodeType {
+  static let breakContextKey = "forloop_break"
```
djbe: Maybe add a bit more of a prefix to these strings to really avoid context collisions? Such as "_internal_forloop_break"?

Or maybe these flags should be stored directly in the context as separate values (and not pollute the dictionary stacks)? That might be a better long-term solution.

Sources/Parser.swift
```diff
@@ -47,7 +48,11 @@ public class TokenParser {
           return nodes
         }
 
-        if let tag = token.components().first {
+        let components = token.components()
+        if var tag = components.first {
+          if tag.hasSuffix(":") && components.count >= 2 {
```
djbe: If we're keeping the label syntax (see question above), we should add a comment here what's happening and why.

djbe:

Sources/ForTag.swift
djbe: Oh I see, no idea what the general style guide is then. Personally I use `?? false` in `return` statements, and sometimes in `if` statements as well.

djbe: Right, so the consensus seems to be either:
- `if components.first?.hasSuffix(":") == true {` (yours)
- `if (components.first ?? "").hasSuffix(":")` (getting rid of the optional)
- `if let firstComponent = components.first, firstComponent.hasSuffix(":") {` (also getting rid of the optional)

ilyapuchka: I meant to use `components[0].hasSuffix(:)`. There always will be at least 1 component, so it's safe. We can also check before that if there is more than one component because that should be the case with labels (`{% label: %}` is not a valid syntax)

djbe: I'm not too big of a fan of using subscript without bounds checks, so I'd add at least an `.isEmpty` check. Or keep the current syntax, my original comment only applied because I didn't notice it was Optional.

ilyapuchka:
Is it something we still want to have or it can be closed?

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Big PR</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe (approved):
Rebased on master, added support for multi-level `forloop` access, and fixed a bug in the tokenizer because of #178. The fix for that last one is "okay-ish" 🤷

PR #178, opened 2017-12-28
Fixed using spaces in filter expressions and variables lists

ilyapuchka:
When parsing filter expression spaces were not trimmed in filter name and parameters. Also this fixes using spaces in any other nodes where smart split by spaces is used, i.e. in for node. It allows use of `a, b` or `a , b`, but not `a ,b`. The same about `|` and `:` separators. So this also fixes #169

### DIFF
--- a/Sources/Tokenizer.swift
+++ b/Sources/Tokenizer.swift
@@ -10,6 +10,18 @@ extension String {
     var singleQuoteCount = 0
     var doubleQuoteCount = 0
 
+    let specialCharacters = ",|:"
+    func appendWord(_ word: String) {
+      if components.count > 0 &&
+        (specialCharacters.characters.contains(components.last!.characters.last!) ||
+          specialCharacters.contains(word))
+      {
+        components[components.count-1] += word
+      } else {
+        components.append(word)
+      }
+    }
+
     for character in self.characters {
       if character == "'" { singleQuoteCount += 1 }
       else if character == "\"" { doubleQuoteCount += 1 }
@@ -19,7 +31,7 @@ extension String {
         if separate != separator {
           word.append(separate)
         } else if singleQuoteCount % 2 == 0 && doubleQuoteCount % 2 == 0 && !word.isEmpty {
-          components.append(word)
+          appendWord(word)
           word = ""
         }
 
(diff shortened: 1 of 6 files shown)


Sources/Tokenizer.swift
```diff
@@ -36,10 +36,37 @@ extension String {
       components.append(word)
     }
 
-    return components
+    return smartJoin(components)
   }
 }
 
+// joins back components around characters used in variables lists and filters
+private func smartJoin(_ components: [String]) -> [String] {
```
yonaskolb: Does this have performance implications?

ilyapuchka: @yonaskolb thanks for pointing it out =) I've reimplemented it, should be faster now.

ilyapuchka:
@kylef can we merge this? I believe this will improve templates readability

Sources/Tokenizer.swift
```diff
@@ -10,6 +10,18 @@ extension String {
     var singleQuoteCount = 0
     var doubleQuoteCount = 0
 
+    let specialCharacters = ",|:"
+    func appendWord(_ word: String) {
+      if components.count > 0 &&
+        (specialCharacters.characters.contains(components.last!.characters.last!) ||
```
AliSoftware: 😱 all those force unwraps 😢

Sources/Tokenizer.swift
```diff
@@ -10,6 +10,18 @@ extension String {
...
+          specialCharacters.contains(word))
+      {
+        components[components.count-1] += word
+      } else {
+        components.append(word)
+      }
+    }
```
AliSoftware: Wouldn't it be more logical to use a `Set<Character>` here which would seem more appropriate and already handle all that append-without-duplicates logic for you already?

Tests/StencilTests/ForNodeSpec.swift
```diff
@@ -111,8 +111,8 @@ func testForNode() {
         try expect(try node.render(context)) == "empty"
     }
 
-    $0.it("can render a filter") {
-      let templateString = "{% for article in ars|default:articles %}" +
+    $0.it("can render a filter with spaces") {
+      let templateString = "{% for article in ars | default: a, b , articles %}" +
```
AliSoftware: Not sure I'm familiar with that default filter accepting 3 arguments?

kylef: It supports variadic fallbacks for the default value. If `a` and `b` are not defined then articles would be there.

Sources/Tokenizer.swift
ilyapuchka: 🤷‍♂️ it's all safe 🙄

Sources/Tokenizer.swift
ilyapuchka: That's not exactly what we are doing here. We are checking if a word is a special character (for the case when it is surrounded by spaces, i.e. in ` | `), or if previous word ends with a special character, in which case we concatenate last and new word (so `ars| default` ends up as `ars|default`).

AliSoftware: Ah right gotcha

AliSoftware:
I'm still not a fan of force-unwrap even especially when we have better constructs to use instead

Sources/Tokenizer.swift
AliSoftware: `if let lastCompChar = components.last?.characters.last, (specialCharacters.contains(lastCompChar) || …)`

ilyapuchka: @AliSoftware done

PR #179, opened 2017-12-29
Range variables

ilyapuchka:
This PR introduces new syntax for `for` loops - `{% for i in a to b %}`. `a to b` will produce a range variable which will render itself as closed range `a...b`. Values can be number literals or context variables, they will be resolved while rendering range variable.

### DIFF
--- a/Sources/Variable.swift
+++ b/Sources/Variable.swift
@@ -144,12 +144,12 @@ public struct RangeVariable: Resolvable {
     guard let to = toResolved.flatMap(toNumber(value:)).flatMap(Int.init) else {
       throw TemplateSyntaxError("to value \(toResolved ?? "nil") is not an int")
     }
-
-    guard from < to else {
-      throw TemplateSyntaxError("from value (\(from)) is not less than to value (\(to))")
+    let range = min(from, to)...max(from, to)
+    if from > to {
+      return Array(range.reversed())
+    } else {
+      return Array(range)
     }
-
-    return CountableClosedRange<Int>(uncheckedBounds: (lower: from, upper: to))
   }
 
 }
--- a/Sources/Variable.swift
+++ b/Sources/Variable.swift
@@ -126,6 +126,34 @@ public func ==(lhs: Variable, rhs: Variable) -> Bool {
   return lhs.variable == rhs.variable
 }
 
+/// A structure used to represet range of to integer values expressed as `from to to`.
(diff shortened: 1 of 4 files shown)


AliSoftware:
I suggest `for i from a to b` instead?

ilyapuchka:
I was thinking about that, but we already use `in` and I didn't want to create another variation for this syntax. After this is merged we may extend it to allow use of `{% if i in a to b %}`, where `in` is an operator and not just separator like in `for`. So it will feel more consistent between those two tags if we use `in`, I think. But I have no problems with doing it with `from`.

Sources/Variable.swift
```diff
@@ -123,6 +123,34 @@ public func ==(lhs: Variable, rhs: Variable) -> Bool {
...
+
+    guard let from = fromResolved.flatMap(toNumber(value:)).flatMap(Int.init) else {
+      throw TemplateSyntaxError("from value \(fromResolved ?? "nil") is not an int")
+    }
+
+    guard let to = toResolved.flatMap(toNumber(value:)).flatMap(Int.init) else {
+      throw TemplateSyntaxError("to value \(toResolved ?? "nil") is not an int")
+    }
+
+    guard from < to else {
```
djbe: Why should `from` be less than `to`? Nothing is stopping us from adding decreasing ranges, and add a check below to if need be, `reverse` the range.

ilyapuchka: Because I used CountableClosedRange that requires that (or is it <= ?). We can of course switch to array here and remove this condition.

ilyapuchka: Regarding reversing this will be hard to do using filter with such syntax: `a to b | reverse` will try to apply filter to b, not to the range. But I guess if you are constructing range manually you wouldn't need to reverse it as you can create it already in a write direction.

djbe: No, I meant more something like:
```
let range = CountableClosedRange<Int>(uncheckedBounds: (lower: from, upper: to)).reversed()
```

djbe: More complete bit of code:
```swift
let lower = min(from, to)
let upper = max(from, to)
var range = CountableClosedRange<Int>(uncheckedBounds: (lower: from, upper: to))
if from > to {
  range = range.reversed()
}
```

AliSoftware:
Btw why use the `init(uncheckedBounds:)`? Can't we use `from...to` directly?

PR #180, opened 2017-12-29
Titlecase filter

ilyapuchka:


AliSoftware:
Is it the same transformation as the `upperFirstLetter` from [StencilSwiftKit](https://github.com/SwiftGen/StencilSwiftKit/blob/master/Documentation/filters-strings.md)?

Note that in the common vocabulary (and Foundation implementation for example), `titlecase` turns the first letter to uppercase… but also turns the other letters to lowercase (while our `upperFirstLetter` only uppercase the first letter letting the others unchanged).

So `"datePicker"|titlecase` would be `"Datepicker"` while `"datePicker"|upperFirstLetter` would be `"DatePicker"`

AliSoftware:
Also be sure to check if `capitalize` isn't the same as your `titlecase` too (which is one provided by Stencil itself iirc)

ilyapuchka:
you're right, capitalise already doing it

PR #182, opened 2017-12-30
Caching rendered blocks content to reuse them in further calls

### LINKED ISSUE #158: Using blocks more than once?
layout.stencil:
```
<!doctype html>
<html>
  <head>
    <title>My Music Collection: {% block pageTitle %}{% endblock %}</title>
  </head>
  <body>
    <h1>My Music Collection</h1>
    <h2>{% block pageTitle %}{% endblock %}</h2>
    {% block pageContent %}{% endblock %}
  </body>
</html>
```

hello.stencil:
```
{% extends "layout.stencil" %}

{% block pageTitle %}Hello!{% endblock %}

{% block pageContent %}
<p>
  Hello, {{ name|default:"World" }}!
</p>
{% endblock %}
```

So my ideal result would be:

```
<!doctype html>
<html>
  <head>
    <title>My Music Collection: Hello!</title>
  </head>
  <body>
    <h1>My Music Collection</h1>
    <h2>Hello!</h2>
    <p>
      Hello, World!
    </p>
  </body>
</html>
```

I think this is a natural way to have a "subtitle" that appears both in the page content and in the `<title>` tag, but it doesn't seem to work - the second `pageTitle` block isn't filled in. So the result looks like:

    <h1>My Music Collection</h1>
    <h2></h2>

Not sure if this counts as a bug. If not, please consider it a feature request.

ilyapuchka:
Resolves #158 
This PR implements caching of rendered blocks context in the top context so that it can be later rendered again using variable expression.

Example:

```
<title>{% block title %}Hello World!{% endblock %}</title>
<h1>{{ block.title }}</h1>
```

Result:

```
<title>Hello World!</title>
<h1>Hello World!</h1>
```

In case of inheritance cached content is updated each time child block is rendered, so when block repeats in the base template its content will be the same as content of the first block invocation, containing all its children contents.

Example:

```
base.html

<title>{% block title %}Hello World!{% endblock %}</title>
<h1>{{ block.title }}</h1>
```

```
child.html

{% extends base.html %}
<title>{% block title %}{{ block.super}} Foo Bar {% endblock %}</title>
```

Result (when rendering child.html):

```
<title>Hello World! Foo Bar</title>
<h1>Hello World! Foo Bar</h1>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@
 - Added support for adding default whitespace trimming behaviour to an environment.  
   [Yonas Kolb](https://github.com/yonaskolb)
   [#287](https://github.com/stencilproject/Stencil/pull/287)
+- Blocks now can be used repeatedly in the template. When block is rendered for the first time its content will be cached and it can be rendered again later using `{{ block.block_name }}`.  
+  [Ilya Puchka](https://github.com/ilyapuchka)
+  [#158](https://github.com/stencilproject/Stencil/issues/158)
+  [#182](https://github.com/stencilproject/Stencil/pull/182)
 
 ### Deprecations
 
--- a/Sources/Stencil/Context.swift
+++ b/Sources/Stencil/Context.swift
@@ -85,4 +85,19 @@ public class Context {
 
     return accumulator
   }
+
+  /// Cache result of block by its name in the context top-level, so that it can be later rendered
+  /// via `{{ block.name }}`
+  ///
+  /// - Parameters:
+  ///   - name: The name of the stored block
+  ///   - content: The block's rendered content
+  public func cacheBlock(_ name: String, content: String) {
+    if var block = dictionaries.first?["block"] as? [String: String] {
+      block[name] = content
+      dictionaries[0]["block"] = block
+    } else {
+      dictionaries.insert(["block": [name: content]], at: 0)
+    }
+  }
 }
--- a/Sources/Stencil/Inheritance.swift
+++ b/Sources/Stencil/Inheritance.swift
@@ -151,6 +151,8 @@ class BlockNode: NodeType {
       }
     }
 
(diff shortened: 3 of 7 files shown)


djbe:
Should we still consider this, as we're planning on merging filters such as `set` from `StencilSwiftKit`?

ilyapuchka:
@djbe I think its a bit different than `set`, though they may be used to achieve similar results

ilyapuchka:
@kylef @djbe @yonaskolb can we move on with that?
@djbe speaking of differences with `set`, `blocks` are rendered when they are defined and they are designed primarily for inheritance.

djbe:
Side Note: `set` is also rendered when it's defined.

I've just read through the related issue #158, where you mention what Jinja does. Seeing as Stencil tries to mirror Jinja where possible, shouldn't we do the same?  I.e. have the block `title` be cached under `self.title`?

ilyapuchka:
@djbe  we can do `self.title` but I think `block.title` is more helpful to recognize what this variable stands for

djbe:
I agree that `block.title` may be clearer, but we're basing ourselves of Jinja, so I think we should mirror that if possible.

@kylef thoughts? You have better knowledge about this.

ilyapuchka:
@djbe I don't think we should prefer mirroring Jinja syntax to better API, where `block.name` is better and we seem to agree on that. We already have quite a few points of divergence with Jinja and it's just not a goal to be canonical Swift implementation on Jinja language 🤷‍♂️

ilyapuchka:
Is this something we want to have or it can be closed?

SwiftGen-Eve:

---

</p>

djbe (approved):
Looks good. I've made the `cacheBlock` function public in case anyone with custom blocks wants to do the same.

PR #183, opened 2017-12-31
Arithmetic operations

ilyapuchka:


ilyapuchka:
Is it something we want to have or it can be closed without pending time resolving conflicts?

akshaynhegde:
Any reason for not having arithmetic expression support? I do have the need for this and would be useful.

PR #188, opened 2018-01-21
Added indent filter

### LINKED ISSUE #95: Question: Indentation control in {% include %}

ilyapuchka:
With this filter it is possible to indent content either of variable or included from another template.

Examples:

```
Indented text: {{ "first line\nsecond line"|indent:2,"\t"}}
```

will render:

```
Indented text: first line
\t\tsecond line
```

With `filter` tag indentation can be also applied to included template content:

```
Included text: {% filter indent:2 %}{% include "text.html" %}{% endfilter %}
```

will render:

```
Included text: first line
  second line
```

Resolves #95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added support for resolving superclass properties for not-NSObject subclasses
 - The `{% for %}` tag can now iterate over tuples, structures and classes via
   their stored properties.
+- Added `indent` filter
 
 ### Bug Fixes
 
--- a/Sources/Extension.swift
+++ b/Sources/Extension.swift
@@ -57,6 +57,7 @@ class DefaultExtension: Extension {
     registerFilter("uppercase", filter: uppercase)
     registerFilter("lowercase", filter: lowercase)
     registerFilter("join", filter: joinFilter)
+    registerFilter("indent", filter: indentFilter)
   }
 }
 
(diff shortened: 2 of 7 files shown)


ilyapuchka:
@kylef all your comments addressed, thanks for taking time to review this =)

kylef:
I think one was missed (documentation): https://github.com/kylef/Stencil/pull/188#pullrequestreview-90352354, otherwise good to go.

ilyapuchka:
@kylef this is ready to be merged

PR #191, opened 2018-01-22
Added valueForKey filter

ilyapuchka:
This PR implements the same functionality as #174 but with less confusing API.


ilyapuchka:
@kylef I've added map for single value in #189 so closing this

PR #192, opened 2018-01-23
Implemented parsing of range literals

ilyapuchka:
This is alternative implementation of #179 that does not introduce new syntax for loops and automatically supports if-in expressions.
If we merge #175 we could even support open ranges, but for now only closed ranges are supported.

Closes #179 
Closes https://github.com/SwiftGen/StencilSwiftKit/issues/74

### DIFF
--- a/Tests/StencilTests/ForNodeSpec.swift
+++ b/Tests/StencilTests/ForNodeSpec.swift
@@ -227,7 +227,7 @@ func testForNode() {
           .block(value: "for i"),
       ]
       let parser = TokenParser(tokens: tokens, environment: Environment())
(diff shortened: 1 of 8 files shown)


Tests/StencilTests/ForNodeSpec.swift
```diff
@@ -173,7 +173,7 @@ func testForNode() {
           .block(value: "for i"),
       ]
       let parser = TokenParser(tokens: tokens, environment: Environment())
-      let error = TemplateSyntaxError("'for' statements should use the following 'for x in y where condition' `for i`.")
+      let error = TemplateSyntaxError("'for' statements should use the following syntax:\n`for x in y where condition`")
```
kylef: `TemplateSyntaxError` message is expected to be a single line

PR #196, opened 2018-02-03
Fixed warnings

ilyapuchka:


ilyapuchka:
@kylef you were mentioning in some comment that we should drop Swift 3. I don't see point to have those conditions then, should we just fix warning and keep Swift 4 version.

kylebshr:
Anything I can do to help move this forward?

djbe:
After we release Stencil 0.12 (which should be soon), we'll drop Swift 3 and get rid of these warnings.

PR #198, opened 2018-02-11
extended context to enable implementation of swiftlike scoping rules …

kaygro:
…(pushLocals function)

added a testcase for this new behavior
modified ForTag to make use of this new mechanism

### DIFF
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -52,6 +52,33 @@ public class Context {
     defer { _ = pop() }
     return try closure()
   }
+	
+	private func pop(_ locals: Set<String>) -> [String: Any?]?{
+		let top = pop() ?? [:]
+		var popped: [String: Any] = [:]
+		//propagate non local preexisting variable values down the stack
+		for (key, value) in top {
+			if !locals.contains(key) && self[key] != nil{
+				self[key] = value
+			}else{
+				popped[key] = value
+			}
+		}
+		if popped.isEmpty{
+			return nil
+		}
+		return popped
+	}
+	
+	//this mimicks typical programming language scoping rules
+	public func pushLocals<Result>(dictionary: [String: Any]? = nil, closure: (() throws -> Result)) rethrows -> Result {
+		let dictionary = dictionary ?? [:]
+		let locals = Set(dictionary.keys)
+		
+		push(dictionary)
+		defer { _ = pop(locals) }
+		return try closure()
+	}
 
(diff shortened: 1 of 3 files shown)


kylef:
Hi @kaygro, I'm having a hard time trying to understand what you are trying to solve that is different from the `push` method on Context which already takes a closure. Could you perhaps elaborate and perhaps provided an example use-case where the regular `push` method does not suffice?

kaygro:
If you want to to implement an extension, which has set tag (as StencilSwiftKit for example),
you want to be able to do mutations inside a for loop, which are visible outside of it.
A resonable implementation of a set tag would do something like 
context[variableName] = value,
but if that happens inside a for loop, it gets popped off afterwards when using regular push.

kaygro:
Exampe Stencil code for this scenario would look like:
{% set foundElem %}false{% endset %}
{% for elem in array %}
  {% if satisfiesInterestingProperty elem %}
    {% set foundElem %}true{% endset %}
  {% endif %}
{% endfor %}
{# if normal push/pop is used, foundElem will always be "false" at this point #}
{% if foundElem == "true" %}
...
{% endif %}

ilyapuchka:
Not sure about this change, feels too invasive, also I think it can be implemented in a simpler way, but I didn't explore it enough to be sure. Also I think current behaviour is the same as in Jinja.
Speaking of example shared, I believe it can be achieved by using filter on array and then checking its count.

```
{{ if array|yourCustomFilterThatChecksForPredicate|count > 0 }}
```

Of course you'll need to implement this filter and register it in environment, so there is no way to do it just with templates.
Another way would be to use `filter` that I've implemented in #189

kaygro:
Well, I think it can be implemnted entriely from within the set node, but that would mean, the implementation of the set node would need to make copies at the various heights of the context stack, where that variable is visible, (for every variable change that is). It would be less cubersome, if we could observe/override the subscript operators of the Context class, but that is not possible due to access beeing public, not open and Context not being a protocol.

djbe:
@kaygro I get what this is trying to achieve (seeing the issues with `set`), but I agree with the others that this might not be the best way to achieve this.

It might be better to have a mechanism in Context to directly modify the stack, so that `set` (for example) could modify a variable outside it's scope. For example

kaygro:
I like that idea. I didn't realize, you could have multiple parameters in subscript operators.

ilyapuchka:
Closing this due to inactivity.

PR #201, opened 2018-03-02
fixes Swift tools version

nhgrif:
I'm not sure if this was working for other people, but we are still finding a conflict between Commander & Stencil.  When trying to resolve my dependencies using Swift tools version 4, this is the dependency graph I get:

└── Stencil<https://github.com/kylef/Stencil.git@0.10.1>
    ├── PathKit<https://github.com/kylef/PathKit.git@0.8.0>
    │   └── Spectre<https://github.com/kylef/Spectre.git@0.7.2>
    └── Spectre<https://github.com/kylef/Spectre.git@0.7.2>

Based on the documentation here: https://swift.org/blog/swift-package-manager-manifest-api-redesign/

> Since Swift 3 Package Manager doesn’t understand the Swift 4 manifest format, it will automatically ignore the Git tags that contain a Swift 4 manifest. So, if a package upgrades to Swift 4 manifest, Swift 3 Package Manager will pick the last tag which contains the Swift 3 manifest. However, the Package Manager in Swift 4 will pick the latest available version, regardless of manifest version.

It sounds like because Stencil's Package specifies tools version 3, the tags in Spectre and PathKit which are specifying version 4 are getting ignored, thus causing a conflict between Stencil and Commander.


nhgrif:
I'll reopen this when I've actually tested that it's working.

nhgrif:
We are able to successfully pull in both commander and stencil now from this branch.

nhgrif:
This PR now includes #196 in order to actually work for Swift 4.

djbe:
Closing this as it's been fixed with 0.12.1, thank you for the PR though!

PR #203, opened 2018-03-22
Added filter to apply dynamic filters

ilyapuchka:
`filter:arg` will resolve its argument as filter expression which will allow to apply dynamic filters. 

The use case for that can be using Sourcery to generated code for Codable protocols and use different key coding strategies applied for different data models. It will allow to specify strategy as one of Stencil filters in a type annotation, i.e. `uppercaseFirstLetter` or `camelCaseToSnakeCase`, or even filter with arguments. 

Currently in template I have to check annotations against hardcoded set of filters to be able to apply them. With this change this template code

```jinja
{% if strategy == "upperFirstLetter" %}{{ variable.name|replace:"_",""|upperFirstLetter }}{% else %}{{ variable.name|replace:"_",""|camelToSnakeCase }}{% endif %}
```

becomes this

```jinja
{{ variable.name|replace:"_",""|filter:strategy }}
```

and allows to use any registered filter without changing anything in template.


AliSoftware:
I'll have to re-check the Stencil docs (on phone rn so not the best device to read them) but I thought Stencil already had a `{% filter xx %}…{% endfilter %}` tag that would also allow you to do dynamic filters already?

ilyapuchka:
@AliSoftware the implementation of filter node only converts the whole node to filter expression applied to the rendered content of the node, it does not resolve the name of the filter as variable. But that's something we can add too! Also I just find filter expression easier to use than filter node and better compassable (chaining vs nesting)

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
   [Ilya Puchka](https://github.com/yonaskolb)
   [#178](https://github.com/stencilproject/Stencil/pull/178)
 
+- Added support for dynamic filter using `filter` filter
```
djbe: If (when?) #221 gets merged, could you fix the format of this entry? Remove the newline above, add a `.` and 2 spaces at the end, and credit yourself + link to this PR.

Sources/Parser.swift
```diff
@@ -71,8 +71,24 @@ public class TokenParser {
     tokens.insert(token, at: 0)
   }
 
+  public func compileFilter(_ token: String) throws -> Resolvable {
```
djbe: Why do these methods (`compileFilter`, `compileExpression` and `compileResolvable`) have to be public?

Note: if they can be internal, don't forget to reset the `Expression` protocol back to internal.

Sources/Parser.swift
```diff
@@ -111,11 +127,15 @@ public class TokenParser {
   }
 
   public func compileFilter(_ token: String) throws -> Resolvable {
```
djbe: This method isn't part of the review, but this is the same comment for these 3 functions: why do these need to be `public`?

Sources/Extension.swift
```diff
@@ -26,6 +26,10 @@ open class Extension {
 
   /// Registers a template filter with the given name
   public func registerFilter(_ name: String, filter: @escaping (Any?, [Any?]) throws -> Any?) {
```
djbe: Should we deprecate this function (and mark it with `@available(*, deprecated`)?

djbe:
Except from the small changelog change, and some small access questions, this LGTM.

djbe:

Sources/Parser.swift
djbe: Ehm, do they need to be used by 3rd party extensions? Do you have a scenario where that'd be necessary?

Sources/Extension.swift
djbe: I'm not saying remove it, just mark it as `deprecated`.

Note: having 2 methods that only differ in the parameters of a callback block is quite annoying for Xcode autocomplete, which doesn't handle this very well (at all).

Sources/Parser.swift
ilyapuchka: I cant think of any now but if I needed it in `filter` then it might be needed in something else complex too. Anyway, I don't see any harm in making these methods public.

Sources/Extension.swift
djbe: Yeah Xcode shouldn't limit our API, but Swift should. And its type inference can sometimes go bonkers on cases like these 🤷‍♂️

Sources/Parser.swift
ilyapuchka: While in custom filters the same methods can be accessed through `context.environment`, in custom nodes they will be accessed through `TokenParser` which `environment` property is prived, so these methods are private to allow custom nodes implementations to use them.

Sources/Extension.swift
ilyapuchka: Maybe, but I didn't came across anything weird when working on that, so I guess it should be file to leave it. Also overload with `context` argument is something most of filters do not need.

ilyapuchka:
@djbe comments addressed

docs/builtins.rst
```diff
@@ -373,3 +373,13 @@ Filter accepts several arguments:
 * indentation character: character to be used for indentation. Default is a space.
 * indent first line: whether first line of output should be indented or not. Default is ``false``.
 
+``filter``
+~~~~~~~~~
+
+Applies filter by its name, provided as an argument.
```
djbe: Applies the filter with the name (provided as argument) to the current expression.

docs/builtins.rst
```diff
@@ -373,3 +373,13 @@ Filter accepts several arguments:
...
+
+.. code-block:: html+django
+
+    {{ string|filter:myfilter }}
+
+This expression will resolve `myfilter` variable and will apply filter with such name to `string` variable.
```
djbe: This expression will resolve the `myfilter` variable, and it will apply the filter with that resolved name to the `string` variable.

Sources/Extension.swift
djbe: Ok on this point.

Sources/Parser.swift
djbe: Honestly, a library should only expose the minimal surface necessary for a developer to use it. If at any point, something is missing, it can be added later on.

Maybe someone else has some thoughts on this point? @AliSoftware?

djbe:
This will need a rebase, based on 0.12. Be careful when rebasing the changelog, as the entry will need to be moved to the right section (under `master`).

Sources/Parser.swift
djbe: Good idea on the docs then. It took me quite a while to understand the codebase a bit, and the documentation in the code is quite sparse.

ilyapuchka:
@djbe I updated docs. We'll need this, in particular making `Expression` type public to add support for `{% call this if condition else that %}` in StencilSwiftKit, unless we move `call` node into Stencil.

djbe:
Looks GTM. How come this branch is till out-of-date after that merge you did one hour ago?

ilyapuchka:
🤷‍♂️

PR #204, opened 2018-03-25
Fixed accessing optional properties through reflection

ilyapuchka:
Resolves https://github.com/kylef/Stencil/pull/110#issuecomment-375998844
When getting descendant that represents optional property casting it to Optional<Any> will just wrap it in another optional, so instead we use a custom protocol that only Optional implements and that provides unwrapped value or returns nil. If property is not optional we will just return its value.


djbe:
That was quick 👍

djbe:
I've tested it locally and can confirm this fixes the issue 🎉

PR #207, opened 2018-04-04
Optimise Scanner performance

trametheka:
Our project has a few larger templates. After going from Swift 3.1 to 4.1, we noticed a huge performance hit and narrowed it down to here. Under one files test case we are seeing improvements from over 8 seconds down to ~0.15 seconds on our Ubuntu server, and ~0.6 seconds down to ~0.15 on macOS. We are seeing about this performance ratio across all files.

This pull also includes the original scan(until: [String]) function rewritten, but this is about twice as slow as the scanForTokenStart approach.

Open to modifying if required.

### DIFF
--- a/Sources/Lexer.swift
+++ b/Sources/Lexer.swift
@@ -7,7 +7,7 @@ struct Lexer {
 
   func createToken(string: String) -> Token {
     func strip() -> String {
-      guard string.characters.count > 4 else { return "" }
+      guard string.count > 4 else { return "" }
       let start = string.index(string.startIndex, offsetBy: 2)
       let end = string.index(string.endIndex, offsetBy: -2)
       let trimmed = String(string[start..<end])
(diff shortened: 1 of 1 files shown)


kylef:
Since this drops Swift 3 support, I'd like to get one final Swift 3 compatible release out before taking a look and benchmarking this. Thanks for the contribution @trametheka just wanted to update you on status.

yonaskolb:
I'd be keen to get this in, and drop Swift 3 support for next release. 
@trametheka could you rebase this?

trametheka:
I'm afraid I no longer use stencil @yonaskolb and don't really have the time at the moment I'm sorry. You're certainly welcome to use the code in this PR if you can rebase it yourself.

Liquidsoul:
Hey @yonaskolb 👋 
#226 may be the rebase of this you wished for 😉

PR #214, opened 2018-04-30
Context param for Include block

yonaskolb:
This is a redo of #95 which only includes the context param which gets passed to the file.
I've also removed the `using` from the param

```
{% include "file.stencil" childContext %}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ### Enhancements
 
-- added an optional second parameter to the `include` tag for passing a sub context to the included file
+- Added an optional second parameter to the `include` tag for passing a sub context to the included file.  
+  [Yonas Kolb](https://github.com/yonaskolb)
+  [#394](https://github.com/stencilproject/Stencil/pull/214)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Master
 
+### Enhancements
+
+- added an optional second parameter to the `include` tag for passing a sub context to the included file
+
 ### Bug Fixes
 
 - Fixed using quote as a filter parameter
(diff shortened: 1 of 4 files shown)


yonaskolb:
@ilyapuchka @AliSoftware Could I get a review on this?

AliSoftware:
@yonaskolb Also, I'm curious as to why you pushed that on your fork — which should not be necessary anymore now that you're a core contributor and have push access — instead of pushing that branch directly on the original repo?

I'm guessing it's because it's a branch that were already open in your end before you became one of the co-owners (actually you don't seem to have accepted my invitation to be part of the team?), but that would still give you some more power (add reviewers, etc), so 😉 

Usually in most projects, CI is disabled for forks for security reasons (would otherwise allow some ill-intended user to expose tokens otherwise, iirc), so pushing a branch on the main repo rather than the fork is generally preferred when possible also for that reason (even if idk if that's the case for Stencil, would have to check on travis).
Also would allow you to get rid of your fork and avoid having to maintain 2 different remotes for the same project and the pain of having to keep them in sync :wink:

yonaskolb:
Thanks @AliSoftware, comments addressed.

Yeah if no one person is managing releases we should add changelog entries in PRs 👍

Yes, the branch is just an older one from my fork.
Any future work will be in this repo 👍

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Master
 
+### Enhancements
+
+- added an optional second parameter to the `include` tag for passing a sub context to the included file
```
djbe: Maybe we should directly start applying the new changelog format? End with a `.  ` (2 spaces), and the author + issue link. See https://github.com/SwiftGen/SwiftGen/blob/master/CHANGELOG.md for examples.

Nitpick: start sentences with an uppercase.

yonaskolb: Cool, done

PR #225, opened 2018-08-24
[WIP] Faster scanner rebase

svanimpe:
This is a rebase of #207.

I've only imported the new code and fixed any warnings, but as it is undocumented, I don't have to time to actually understand it :D

@djbe Can you take a look at this? The unit tests are failing and I suspect it's because the new code overwrites parts of ffe8f9d (or the new implementation just behaves differently, or has bugs).

### DIFF
--- a/Sources/Lexer.swift
+++ b/Sources/Lexer.swift
@@ -98,9 +98,10 @@ class Scanner {
     for char in content.unicodeScalars {
       if char == first {
         //Check the rest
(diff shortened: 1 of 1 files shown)


Liquidsoul:
Hey @svanimpe 👋 
It seems we've been working on the same rebase #226 😅 
I had more time to integrate the upstream changes and fix the tests. Maybe you could give it a quick look? 👀

svanimpe:
OK, this one can be closed then!

Issue #224, opened 2018-08-19
Critical performance issue


svanimpe:
After refactoring my app, I'm seeing extreme rendering times, as bad as 14s sometimes. I'm not entirely sure what causes this, but I wanted to report it already, as I believe the issue lies with Stencil.

Here's what I have so far:

- I'm using Stencil 0.11.
- I didn't notice any issue in development. On macOS, 0.11 works fine and render times are < 100ms usually.
- When deploying the same code to IBM Cloud, the render times become horrible (1-15s). This leads me to believe it's a Linux-specific issue. I'm not sure about that however, as I haven't been able to reproduce it on Linux via Docker, only when deploying to cloud.
- I did not have this issue before when I was using 0.10.1.
- I was able to reproduce the issue to a lesser extent on macOS as well. If I use the latest master, my render times jump over 1s as well.
- I've added logging all throughout my handlers as well as Kitura and didn't notice any issues there. Almost all of the delay is caused by Stencil. I went as far as `Template.render` but haven't narrowed it down further yet.
- All my templates uses inheritance. I'm having other issues with inheritance on master as well, so perhaps it related?

I would really appreciate any help in tracking down and resolving this issue as I need to release this app in the next few weeks.

I've also tried reverting back to 0.10, but that's giving me a compilation error on Linux I've yet to resolve:

`PathKit.swift:299:11: error: cannot convert value of type 'ObjCBool' to expected argument type 'Bool'`

svanimpe:
I've logged every render separately and came up with the following results:

```
 TextNode in 0ms
 TextNode in 0ms
 TextNode in 0ms
 TextNode in 0ms
 VariableNode in 0ms
 TextNode in 0ms
 IfNode in 1ms
 TextNode in 0ms
 BlockNode in 62ms
 TextNode in 0ms
 VariableNode in 19ms
 TextNode in 0ms
 VariableNode in 19ms
 TextNode in 0ms
 TextNode in 0ms
 VariableNode in 0ms
 TextNode in 0ms
 BlockNode in 3ms
 TextNode in 0ms
 BlockNode in 0ms
 TextNode in 0ms
 IfNode in 0ms
 TextNode in 0ms
 TextNode in 0ms
 IfNode in 2ms
 TextNode in 0ms
 TextNode in 0ms
 IfNode in 0ms
 TextNode in 0ms
 TextNode in 0ms
 TextNode in 0ms
 IfNode in 0ms
 TextNode in 0ms
[...]

PR #226, opened 2018-08-24
Optimise Scanner performance

Liquidsoul:
This is a rebase of #207

**✅ What I did:**
- squash of the commits
- integrate the upstream changes about range reporting to fix the tests during the rebase process
- remove Swift 3 from the build matrix

**❌What I did not do:**
- benchmark if there is an actual performance improvement

### DIFF
--- a/Sources/Lexer.swift
+++ b/Sources/Lexer.swift
@@ -7,6 +7,13 @@ struct Lexer {
   let templateString: String
   let lines: [Line]
 
+  private static let tokenChars: [Unicode.Scalar] = ["{", "%", "#"]
+  private static let tokenCharMap: [Unicode.Scalar: Unicode.Scalar] = [
+    "{": "}",
+    "%": "%",
+    "#": "#"
+  ]
+
   init(templateName: String? = nil, templateString: String) {
     self.templateName = templateName
     self.templateString = templateString
@@ -20,9 +27,7 @@ struct Lexer {
   func createToken(string: String, at range: Range<String.Index>) -> Token {
     func strip() -> String {
       guard string.count > 4 else { return "" }
-      let start = string.index(string.startIndex, offsetBy: 2)
-      let end = string.index(string.endIndex, offsetBy: -2)
-      let trimmed = String(string[start..<end])
+      let trimmed = String(string.dropFirst(2).dropLast(2))
         .components(separatedBy: "\n")
         .filter({ !$0.isEmpty })
         .map({ $0.trim(character: " ") })
@@ -55,21 +60,14 @@ struct Lexer {
     var tokens: [Token] = []
 
     let scanner = Scanner(templateString)
-
-    let map = [
-      "{{": "}}",
-      "{%": "%}",
-      "{#": "#}",
-      ]
-
     while !scanner.isEmpty {
-      if let text = scanner.scan(until: ["{{", "{%", "{#"]) {
-        if !text.1.isEmpty {
-          tokens.append(createToken(string: text.1, at: scanner.range))
+      if let (char, text) = scanner.scanForTokenStart(Lexer.tokenChars) {
+        if !text.isEmpty {
+          tokens.append(createToken(string: text, at: scanner.range))
         }
 
-        let end = map[text.0]!
-        let result = scanner.scan(until: end, returnUntil: true)
+        guard let end = Lexer.tokenCharMap[char] else { continue }
+        let result = scanner.scanForTokenEnd(end)
         tokens.append(createToken(string: result, at: scanner.range))
       } else {
         tokens.append(createToken(string: scanner.content, at: scanner.range))
@@ -95,6 +93,9 @@ class Scanner {
   var content: String
   var range: Range<String.Index>
 
+  private static let tokenStartDelimiter: Unicode.Scalar = "{"
+  private static let tokenEndDelimiter: Unicode.Scalar = "}"
+
   init(_ content: String) {
     self.originalContent = content
     self.content = content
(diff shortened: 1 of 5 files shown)


Liquidsoul:
@yonaskolb PR rebased and entry added to the changelog

yonaskolb:
So I did some benchmarking using SwagGen as a test harness and using XCTest to `measure` just the stencil file generation

Results in seconds:
- `0.11.0` - 0.24
- #207 - 0.15
- `0.12.0` - 5.9
- #228 (this pr) - 4.5

So while this PR improves things slightly it seems something was added after the original #207 that slowed down things even more.

After some more investigation it seems #167 caused the slowdown
before: 0.25
after: 6.1

yonaskolb:
@ilyapuchka as the author of #167 do you have any other insights here?

ilyapuchka:
@yonaskolb I suppose something around here might have caused performance issues https://github.com/stencilproject/Stencil/pull/167/files#diff-91493e167a3f6240a7cc916d9504d439R25

ilyapuchka:
I noticed recently that parsing things like `func some() {{% some tag %}` fails due to usage of `{{%`. I see that there are some changes to the parsing of starting tokens here and wonder if this is still the case with this change and, if so, if it can be addressed @Liquidsoul

Liquidsoul:
I've tried to tokenize `func some() {{% if %}` using this updated `Lexer` and the existing one. Here is the results:
- the new lexer does not handle this well and crashes even.
- the old lexer fails to parse the `if` tag but does not stop the execution.

So the new lexer does solve the issue and introduces a new problem.
Instead of failing to capture the tag as in the old one the new lexer crashes.
I see that as a blocking regression.

The fact that we did not have a testing context showing the improvement in #207 makes it hard to see this as a working optimisation. And since the original author does not seem to use Stencil anymore maybe we should just drop this entirely, see if this reported again and just address the existing performance issue.

What do you think?

ilyapuchka:
@yonaskolb can you create a PR with your performance tests?

djbe:
- #228 has been merged so this should be next.
- #228 introduced a performance loss in this commit: 93ccc56
-  Investigate if String/Substring is the performance issue, and how to solve this.

djbe:
@yonaskolb You've mentioned a few times that you timed the tests. Are you just running them locally and checking how long yourself (with `time` diff or something), or do you have an actual test case we could run on CI?

yonaskolb:
I've just been timing locally using an XCTest in SwagGen while pointing to specific Stencil commit. I don't know of a way to run performance tests on CI as to fail on performance regressions you need to set a baseline and that won't be stable across different machines

djbe:
I don't think we need a test that "fails", more a test that might report the change, either in CI, or here in the PR. For example, SwiftLint reports performance changes in the PR: https://github.com/realm/SwiftLint/pull/2396.

djbe:
Now that whole process would be a separate PR, but just for this commit, so we can test locally, a simple test that runs a few times, and logs the average time would be nice.

ilyapuchka:
@Liquidsoul I created #235 for mentioned issue

djbe:
So, going through the comments again:
- #167 (error log improvements) added a massive slowdown, we should try to figure out why and solve this in a separate PR.
- This PR seems to have a net positive effect, similar to the old #207.
- Both this and the old #207 PR add a bug with `{{%`, this needs to be solved before merging this.
- We need to decide what `{{% ...` should do, and add a test for it (see #235).
- We need some form of performance tests to keep track of things.

Liquidsoul:
@djbe the `{{%` error already exists in the current code base, this is not something that's introduced by this change.
However, as I said [here](https://github.com/stencilproject/Stencil/pull/226#issuecomment-419452265), this PR introduces a crash instead of a parsing failure when there is a syntax error in the stencil file.
So as I said, if we want to implement this optimisation, we may wish to rewrite this to ensure that no regressions are introduced.

djbe:
I get what you mean, but a crash is an order of magnitude worse. And like I said, it’s not really an error unless you define what the syntax should be for such edge cases.

djbe:
Right, so I completely missed #230, which fixed the performance issues from #167 (error improvements).

I've rebased again, and rewritten the PR a bit to squeeze as much performance as I can. I've also fixed the crash, so parsing will continue when it encounters `{{%`. One thing of note is that the `range = ...` calculations (added with the error improvements) also impact performance a bit, so I've tried minimising those. Funnily enough, switching to `Substrings` where possible has a negative impact on performance 🤔.

Using @Liquidsoul's test method (SwaGen) this goes from (an average of) 0.656s to (an average of) 0.419s.

yonaskolb:
The new test will have to be updated to the Spectre 9.0 format

Sources/Lexer.swift
```diff
@@ -105,64 +108,49 @@ class Scanner {
...
     range = range.upperBound..<range.upperBound
-    while index != content.endIndex {
-      let substring = String(content[index...])
-      for string in until {
-        if substring.hasPrefix(string) {
-          let result = String(content[..<index])
-          content = substring
-          return (string, result)
-        }
+    for (index, char) in zip(0..., content.unicodeScalars) {
```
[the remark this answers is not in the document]
djbe: It's bad practice in Swift to use `.enumerated()` just to be able to iterate with an index.

djbe:
@yonaskolb What do you mean the new format? The only difference that I saw in that PR, was just an `XCTestCase` class added in each file, which this PR already has since it's been rebased on `master`.

Sources/Lexer.swift
```diff
@@ -105,64 +108,49 @@ class Scanner {
...
-
-        content = substring
+    for (index, char) in zip(0..., content.unicodeScalars) {
+      if foundChar && char == Scanner.tokenEndDelimiter {
+        let startIndex = content.index(content.startIndex, offsetBy: index)
+        let result = String(content[..<startIndex])
+        content = String(content[content.index(after: startIndex)...])
+        range = range.upperBound..<originalContent.index(range.upperBound, offsetBy: index + 1)
         return result
+      } else if char == tokenChar {
```
ilyapuchka:
```
else  {
  foundChar = char == tokenChar
}
```

Sources/Lexer.swift
```diff
@@ -105,64 +108,49 @@ class Scanner {
...
-          return result + until
-        }
-
-        content = substring
+    for (index, char) in zip(0..., content.unicodeScalars) {
+      if foundChar && char == Scanner.tokenEndDelimiter {
+        let startIndex = content.index(content.startIndex, offsetBy: index)
+        let result = String(content[..<startIndex])
+        content = String(content[content.index(after: startIndex)...])
+        range = range.upperBound..<originalContent.index(range.upperBound, offsetBy: index + 1)
```
ilyapuchka: why `content` and `range` are needed when `result` is returned right after they are calculated? I can't see them being used.

ilyapuchka: can use `...` instead of `+ 1` ?

Sources/Lexer.swift
```diff
@@ -105,64 +108,49 @@ class Scanner {
...
-          content = String(substring[until.endIndex...])
-          return result + until
-        }
-
-        content = substring
+    for (index, char) in zip(0..., content.unicodeScalars) {
+      if foundChar && char == Scanner.tokenEndDelimiter {
+        let startIndex = content.index(content.startIndex, offsetBy: index)
+        let result = String(content[..<startIndex])
+        content = String(content[content.index(after: startIndex)...])
```
ilyapuchka: why not just use `prefix`/`suffix`?

Sources/Lexer.swift
```diff
@@ -105,64 +108,49 @@ class Scanner {
...
-          return (string, result)
-        }
+    for (index, char) in zip(0..., content.unicodeScalars) {
+      if foundBrace && tokenChars.contains(char) {
+        let startIndex = content.index(content.startIndex, offsetBy: index - 1)
+        let result = String(content[..<startIndex])
+        content = String(content[startIndex...])
+        range = range.upperBound..<originalContent.index(range.upperBound, offsetBy: index - 1)
+        return (char, result)
+      } else if char == Scanner.tokenStartDelimiter {
```
ilyapuchka:
```
else {
  foundBrace = char == Scanner.tokenStartDelimiter
}
```

Sources/Lexer.swift
djbe: Yeah, it took me a while before I grokked the whole thing 😄

ilyapuchka:
Can we add a performance test?

Sources/Lexer.swift
djbe: I'll see if it impacts performance.

djbe:
I don't know, you can only really test it with really (really) large templates, or a whole bunch of templates. Maybe leave that for a separate PR?

ilyapuchka:
we could just add whatever project you are using right now to measure improvement and use its templates in a test?

djbe:
I'm using SwaGen (what @yonaskolb was using), it's a whole project, with a whole bunch of templates, and the test isn't perfect as it's a full render, not just the lexer.

Sources/Lexer.swift
djbe: Using `...` gives a `ClosedRange` instead of a `Range`, so we can't use that unfortunately.

djbe:
I've added a performance test based on the template in #224, repeated a few times 😆.

There's no reporting yet, we'll leave that for a different PR (Danger).

djbe:
@yonaskolb @ilyapuchka any final thoughts? I'd be ideal to have this in for 0.13.

Sources/Lexer.swift
```diff
@@ -20,9 +27,7 @@ struct Lexer {
   func createToken(string: String, at range: Range<String.Index>) -> Token {
```
AliSoftware: I admit it's quite hard to review such code (on which I haven't been really working myself, so this review is quite out of context too) without any doc-comments on what the function is supposed to do.

At least a small doc-comment telling the preconditions (= it's expecting a `string` which already starts with `{` + one of the `tokenChars` and ends with the corresponding `tokenCharMap` + `}`) and what the `range` is supposed to mean here, would help.

_I mean… I'm assuming (from reading the implementation) that the check of the `string` starting with `{{/{%/{#`and ending with the corresponding closing sequence… is done in a separate method upstream, right? But I'm just guessing at that point_

djbe: Yeah the checks happen in `Scanner.scanForTokenStart` and `Scanner.scanForTokenEnd`

This method receives a token's content, and creates a corresponding `.variable`, `.block`, `comment` or `.text` token. The range is used for the improved error functionality, so that the parser can point to where in the file the issue is.

PR #228, opened 2018-09-01
Swift 4.1

yonaskolb:
This updates the codebase to Swift 4.1 and drops Swift 3 support

Warnings were not cleaned up in `Lexer.swift` as https://github.com/stencilproject/Stencil/pull/226 will take care of that


Package.swift
```diff
@@ -1,10 +1,22 @@
-// swift-tools-version:3.1
+// swift-tools-version:4.0
 import PackageDescription
 
 let package = Package(
   name: "Stencil",
+  products: [
+        .library(name: "Stencil", targets: ["Stencil"]),
+    ],
```
AliSoftware: OCD-me speaking: seems like there's inconsistent indentation here compared to the rest of the file (tab vs space probably?)

yonaskolb: Fixed 👍

yonaskolb:
Comments addressed and changelog entry added

CHANGELOG.md
```diff
@@ -8,7 +8,9 @@
 
 ### Breaking Changes
 
- _None_
+- Now requires Swift 4.1.
```
kylef: Now requires Swift 4.1 or newer*

djbe:
Would you mind adding the Swift 4 `Lexer.swift` changes to this PR?

It seems we're going to need to do some more testing, to compare #225 and #226, maybe even add some sort of test to keep track of the performance (so it doesn't tank again), and it's easier if that PR doesn't also have changes related to the Swift migration.

CHANGELOG.md
yonaskolb: Done

yonaskolb:
I'd rather leave `Lexer.swift` alone. To fix the warnings means redoing the substring accessing anyway. I'd rather not cause conflicts in the other performance PRs.

yonaskolb:
I just timed this with my tests. It looks like the last commit took it from `0.48` to `1.07`...

djbe:
Could it be the `String(` initializers? We may want to try to keep using `SubString`s, as it avoids re-allocation/copy operations.

yonaskolb:
I tried to keep things as substrings, but things like `hasPrefix` weren't available on linux

djbe:
And if we define our own functions specifically for Linux?

Note: We should move this discussion to #226.

PR #230, opened 2018-09-08
[WIP] Performance improvements

ilyapuchka:
Trying to address performance issues introduced by #167 

- do not parse lines for each token, do that once when lexer is created


ilyapuchka:
@yonaskolb if you have a chance could you please try this branch with your benchmark or add measurement tests if you have written any already?

yonaskolb:
@ilyapuchka I ran the same tests again and got great results!
before: 6.42
after: 0.45

Nice work! 🎉

ilyapuchka:
@yonaskolb thanks! seems acceptable so we can merge it then and someone makes a patch release?

PR #233, opened 2018-09-10
Add Package.swift for Swift 4.2

keith:
This allows Stencil to be compiled with Swift 4.2

We should wait for new releases including https://github.com/kylef/PathKit/pull/56 and https://github.com/kylef/Spectre/pull/37

Also once Swift 4.2 is officially released we can add a CI job for it.

### DIFF
--- a/Package@swift-4.2.swift
+++ b/Package@swift-4.2.swift
@@ -1,0 +1,23 @@
+// swift-tools-version:4.2
+import PackageDescription
+
+let package = Package(
+  name: "Stencil",
+  products: [
+    .library(name: "Stencil", targets: ["Stencil"]),
+  ],
+  dependencies: [
+    .package(url: "https://github.com/kylef/PathKit.git", from: "0.9.0"),
+    .package(url: "https://github.com/kylef/Spectre.git", from: "0.8.0"),
+  ],
+  targets: [
+    .target(name: "Stencil", dependencies: [
+      "PathKit",
+    ], path: "Sources"),
+    .testTarget(name: "StencilTests", dependencies: [
+      "Stencil",
+      "Spectre",
+    ])
+  ],
+  swiftLanguageVersions: [.v4, .v4_2]
+)


keith:
It would be nice once this was merged if there was a new release with this for upstream libraries

djbe:
Do we really need a separate package.swift for this? Can't we just add the supported versions to the existing one?

keith:
The format for supported versions changed from an array of integers to an array of a new set of enum values with the 4.2 compiler.

keith:
There's a possibility we could use a compiler version check in the Package.swift to set the right type, I can try that path if you'd like

djbe:
I don't know if all that's necessary.

With the old format, we can only say `[4]` right? But can you confirm that, as it stands right now, Stencil doesn't work with Swift 4.2? Or is this just to have the nicer syntax with the enums?

keith:
Right now if a package depends on Stencil, it will build it with the 4.0 mode of the 4.2 compiler. By adding this file and these supported versions, it would build it with the 4.2 mode of the 4.2 compiler. Users can continue to use it in the 4.0 mode successfully, but at some point it makes sense to support 4.2 like this

djbe:
Right, but we only just added 4.1 support, partially (see #226). I agree that, when we migrate to 4.2 (if there's anything to migrate), we should update our `podspec` and `Package.swift` to reflect this.

For now, I think this is too early.

djbe:
Whoa, I wouldn't close it, let's keep it for when we're ready. I'm pretty sure that once we figure #226 out, we can do the 4.2 migration. 

(I can't seem to reopen this)

keith:
Ah, I thought you meant longer term. I deleted my fork which is why you can't reopen

djbe:
Going through #226 again, I'm pretty sure that full Swift 4.1 migration is going to take a while...

Well when we do the migration we'll base the package.swift on this one.

PR #234, opened 2018-09-11
Check for property via selector before using value(forKey:)

### LINKED ISSUE #231: SIGABRT crash with key value coding
Hi! I noticed that if you reference a field for a variable that doesn't exist, you get a crash with SIGABRT. For example this template render:

```swift
try Template(templateString: "{{ user.name }}").render(["user": "Blob"])
```

will crash.

To fix I believe all you have to do is guard this `value(forKey:)` call:

https://github.com/stencilproject/Stencil/blob/00e71c1b4dc2ec6c0f87841a10e34299e6059fcc/Sources/Variable.swift#L107

with this:

```swift
        if object.responds(to: Selector(bit)) {
          current = object.value(forKey: bit)
        }
```

Does the current code work as expected, or is it worth fixing?

Thanks!

ilyapuchka:
Resolves #231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug Fixes
 
-- Now accessing undefined keys in NSObject does not cause runtime crash and instead renders empty string.
+- Now accessing undefined keys in NSObject does not cause runtime crash and instead renders empty string.  
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#234](https://github.com/stencilproject/Stencil/pull/234)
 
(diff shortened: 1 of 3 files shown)


djbe:
Ideally we'd be able to log some sort of warning instead of silently continuing. More often than not, I prefer the compiler/runtime to tell me if I did something wrong instead of just doing it's own thing.

What's the general consensus on this? Error out (current behaviour), or silently continue? @AliSoftware @kylef @yonaskolb

ilyapuchka:
This is consistent with current behavior for undefined properties. If we want to revisit it, which I agree we, I think better to do it separately.

mbrandonw:
Thanks for taking care of that so quickly @ilyapuchka!

PR #240, opened 2018-09-20
Deterministic `for` loops for dictionaries.

djbe:
⚠️ Note: based on #239, so merge that first!

With Swift 4.2, hashing isn't consistent between runs anymore. A consequence of this is, if we iterate over a dictionary, the output won't always be the same, which will be unexpected for a user.

With this PR we iterate over the sorted keys of a dictionary, ensuring consistent output across runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Now accessing undefined keys in NSObject does not cause runtime crash and instead renders empty string.  
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#234](https://github.com/stencilproject/Stencil/pull/234)
+- `for` tag: When iterating over a dictionary the keys will now always be sorted to ensure consistent output generation.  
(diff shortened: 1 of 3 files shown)


ilyapuchka:
I think it's fine though I don't think this was mentioned anywhere in docs that on dictionaries the order sill be Deterministic. But this will preserve current behaviour at least.

djbe:
I think end users would expect consistent behaviour, no matter what the underlying (Swift) behaviour might be?

djbe:
And with the new swift hashing, we're forced to do something ourselves in this case.

ilyapuchka:
🤷‍♂️ depends on user understanding of nature of such types =) for some deterministic order in swift < 4.2 could be a surprise too =)

djbe:
@ilyapuchka I inadvertently merged this into the wrong branch 🤦‍♂️, so we should continue our discussion in #242.

PR #242, opened 2018-09-20
Deterministic `for` loops for dictionaries

djbe:
Note: I was a bit too quick and merged #240 into the wrong branch, and github won't allow me to fix things without a new PR 🤷‍♂️ 

With Swift 4.2, hashing isn't consistent between runs anymore. A consequence of this is, if we iterate over a dictionary, the output won't always be the same, which will be unexpected for a user.

With this PR we iterate over the sorted keys of a dictionary, ensuring consistent output across runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Now accessing undefined keys in NSObject does not cause runtime crash and instead renders empty string.  
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#234](https://github.com/stencilproject/Stencil/pull/234)
+- `for` tag: When iterating over a dictionary the keys will now always be sorted (in an ascending order) to ensure consistent output generation.  
+  [David Jennes](https://github.com/djbe)
+  [#240](https://github.com/stencilproject/Stencil/pull/240)
 
 ### Breaking Changes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Now accessing undefined keys in NSObject does not cause runtime crash and instead renders empty string.  
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#234](https://github.com/stencilproject/Stencil/pull/234)
+- `for` tag: When iterating over a dictionary the keys will now always be sorted to ensure consistent output generation.
```
ilyapuchka: [nit] should mention sorting order probably

djbe:
So, small discussion point from #240 is that the current Stencil behaviour was, until now, undefined.

So let's define it then. An important note should be that the behaviour should be consistent, no matter the swift version you compile Stencil with (4 vs 4.2).
- When compiled with Swift 4, the output was always in the same order (but unsorted)
- When compiled with Swift 4.2, the output is random (and unsorted)
- This PR ensures that the output is always in the same order, and to do so, it's sorted (ascending).

CHANGELOG.md
djbe: "... sorted (in an ascending order) ..."?

ilyapuchka: yep

PR #245, opened 2018-09-22
Allow using collection accessors on strings

ilyapuchka:


Sources/Variable.swift
```diff
@@ -100,6 +100,20 @@ public struct Variable : Equatable, Resolvable {
         } else if bit == "count" {
           current = array.count
         }
+      } else if let string = current as? String {
```
djbe: Why repeat this whole `else if` branch, if you can just use the if branch above?

ilyapuchka: I tried to reuse it but you can't interchangeably use `[Any]` and `String`, using available collections type erasures causes other issues like accessing via subscript

ilyapuchka: We could use `string.character` but it will give deprecation warning

djbe: How about extracting the branch into something like `func resolve<T: RandomAccessCollection>(items: T, bit: String) {`?

ilyapuchka: yes, this will probably work

djbe: Quick test in playgrounds:
```swift
func resolve<T: Collection>(items: T, bit: String) -> Any? {
	if let index = Int(bit) {
		if index >= 0 && index < items.count {
			return items[items.index(items.startIndex, offsetBy: index)]
		} else {
			return nil
		}
	} else if bit == "first" {
		return items.first
	} else if bit == "last" {
		return items[items.index(items.endIndex, offsetBy: -1)]
	} else if bit == "count" {
		return items.count
	} else {
		return nil
	}
}
```

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#243](https://github.com/stencilproject/Stencil/pull/243)
 
+- Now you can access string characters by index or get string length the same was as if it was an array, i.e. `{{ 'string'.first }}`, `{{ 'string'.last }}`, `{{ 'string'.1 }}`, `{{ 'string'.count }}`
```
djbe: Missing the `.  ` at the end here.

Sources/Variable.swift
```diff
@@ -87,19 +105,9 @@ public struct Variable : Equatable, Resolvable {
...
-            current = nil
-          }
-        } else if bit == "first" {
-          current = array.first
-        } else if bit == "last" {
-          current = array.last
-        } else if bit == "count" {
-          current = array.count
-        }
+        current = resolveCollection(array, bit: bit)
```
djbe: Would it be worth changing the test to just `if let collection = current as? Collection {`, or would that be too broad of a cast?

ilyapuchka: `Collection` is a generic protocol so can't be used like that.

djbe: Right 🤦‍♂️

Sources/Variable.swift
```diff
@@ -75,6 +75,24 @@ public struct Variable : Equatable, Resolvable {
       return bool
     }
 
+    func resolveCollection<T: Collection>(_ collection: T, bit: String) -> Any? {
```
djbe: I'd actually move this out of the body of `resolve(`, I'm preparing another PR that adds swiftlint, and this will only worsen the situation. Maybe just a `private func` right after this function?

Sources/Variable.swift
```diff
@@ -136,6 +118,24 @@ public struct Variable : Equatable, Resolvable {
   }
 }
 
+fileprivate func resolveCollection<T: Collection>(_ collection: T, bit: String) -> Any? {
```
djbe: We're in the same type (`Variable`), so `private` is enough, `fileprivate` is for accessing from other types.

ilyapuchka:
@djbe updated

djbe:
In the docs, I'd change (https://github.com/stencilproject/Stencil/blob/master/docs/templates.rst#variables):
> Array lookup (first, last, count, index)

To
> Collection lookup (first, last, count, index) for `Array` and `String`

PR #246, opened 2018-09-22
Dynamic member lookup

### LINKED ISSUE #219: Computed properties on objects appear to not be supported
If an object or struct is passed to a template, you can generally access the properties on the object/struct by using `{{ objectName.propertyName }}`. However, this appears not to work with computed properties.

In other words, with a class like:
```swift
class foo {
  var bar: String = {
    return "baz"
  }
}
```

If you instantiate a `foo`, put it in the context dictionary, and then try to print `{{ foo.bar }}` in the template, nothing prints.

In my actual code, I didn't want my `bar` value to be an actual normal property calculated when init-ing the struct, so I ended up working around it by implementing `bar` as a Stencil filter.

ilyapuchka:
This is similar to `@dynamicMemberLookup` in Swift. Would be ideal if there was a way to check if the type supports this in runtime but it's not possible, so had to introduce a custom protocol for that.
Alternatively, we could have a protocol `StencilContextConvertible` that will require to expose dictionary to be used in template context, but that will require to add all properties in such dictionary, even those accessible through reflection, instead of only those which are not accessible this way.
Unfortunately, this does not work out of the box (by just adding `DynamicMemberLookup` conformance) with `@dynamicMemberLookup` unless type defines subscript with `Any?` return type.
Resolves #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ _None_
 - Made the `Template.render(_:)` method (that accepts a `Context`) public.  
   [David Jennes](https://github.com/djbe)
   [#322](https://github.com/stencilproject/Stencil/pull/322)
+- Enable dynamic member lookup using a new `DynamicMemberLookup` protocol. Conform your own types to this protocol to support dynamic member from with contexts.  
+  [Ilya Puchka](https://github.com/ilyapuchka)
+  [#219](https://github.com/stencilproject/Stencil/issues/219)
+  [#246](https://github.com/stencilproject/Stencil/pull/246)
 
 ### Deprecations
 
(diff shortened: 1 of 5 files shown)


schwa:
This is a really useful addition to stencil (that I'm currently using) that really ought to be merged in. Very handy.

schwa:
Presuming this 4 year old PR isn't going to get merged?

SwiftGen-Eve:

---

</p>

djbe (approved):
Unfortunately there's yet to be a mechanism in Swift to check (at runtime) if a type supports this. So the marker protocol is the next best thing.

PR #247, opened 2018-09-22
Update to Spectre 0.9.0

ilyapuchka:
It has some improvements to Xcode integration. Also used `swift test --generate-linuxmain`. And `Package.resolved` is not ignored anymore to ensure dependencies are always in sync.


ilyapuchka:
@djbe as you are on setting up swiftlint for this project, maybe you can add danger too for this kind of stuff? I also wonder if there is a plugin for checking a spelling in doc comments...

djbe:
I can take a look at it, although we haven't even set it up for SwiftGen yet. @AliSoftware started with it quite a while ago with https://github.com/SwiftGen/templates/pull/12, but we haven't taken the time yet to finish things up.

But first let's add SwiftLint, and continue from there.

AliSoftware:
I'd totally love to add a Danger bot for those kind of things! Just never found the time (you know, real life and not enough free time and all).

I don't really like that for now we have to remind that every time in person and block PRs for that (to avoid accidentally merging them without the CHANGELOG), as it might feel quite harsh even if necessary, so a Danger bot would make things way better indeed!

Would love a PR for that ; unfortunately I haven't found a CHANGELOG plugin which matches our CHANGELOG specific style but at least we could warn if it hasn't changed at all at first, and refine that later to also check that the format (esp. fill stop then two spaces) is right, which is the main thing missing from existing Danger plugins.

PR #249, opened 2018-09-24
SwiftLint integration

djbe:
Note: I recommend you view these changes either by checking the "Hide whitespace changes" in GitHub, or in a local git client and hiding whitespace changes.

This adds the `swiftlint` configuration files to the project, and the necessary steps to CI.
- The changes in `Sources/` are pretty minimal, most done by `swiftlint autocorrect`, and the rest are tiny renames/moving code around.
- By contrast, the changes in `Tests/` are quite extensive. Nothing really functional, more in the sense of reorganisation. The problem is that the old tests were all one (insanely) massive function per file, I've had to split these up into smaller functions. A remaining issue is that we may want to further split up the tests into smaller files.

The advantage of the last item, is much nicer Xcode integration for some tests, as some tests are a single `test` function now, showing up in the "Test Navigator". If the rest of the team agrees with this approach, I can apply it to the rest of the test cases. It's also a partial application of https://github.com/orgs/stencilproject/teams/coreteam/discussions/5.


Package.swift
```diff
@@ -4,19 +4,19 @@ import PackageDescription
 let package = Package(
   name: "Stencil",
   products: [
-    .library(name: "Stencil", targets: ["Stencil"]),
+    .library(name: "Stencil", targets: ["Stencil"])
   ],
   dependencies: [
     .package(url: "https://github.com/kylef/PathKit.git", from: "0.9.0"),
-    .package(url: "https://github.com/kylef/Spectre.git", from: "0.9.0"),
```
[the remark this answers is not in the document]
djbe: I'd rather avoid a bunch of copy/paste? More often than not, it leads to mistakes.

Tests/StencilTests/EnvironmentSpec.swift
```diff
@@ -1,342 +1,392 @@
...
+    reason: String,
+    token: String,
+    file: String = #file,
+    line: Int = #line,
+    function: String = #function
+  ) throws {
+    let expectedError = expectedSyntaxError(token: token, template: template, description: reason)
+
+    let error = try expect(
+      self.environment.render(template: self.template, context: ["names": ["Bob", "Alice"], "name": "Bob"]),
```
ilyapuchka: extract context in a local variable and also move template and context parameters on separate lines?

ilyapuchka: I'd prefer to keep this function in a single line though, same as expectation later.

djbe: > extract context in a local variable
Could you clarify this bit?

ilyapuchka:
```
let context = ["names": ["Bob", "Alice"], "name": "Bob"])
self.environment.render(template: template, context: context, file: file, line: line, function: function)
```

Tests/StencilTests/EnvironmentSpec.swift
```diff
@@ -1,342 +1,392 @@
...
+
+    let path = Path(#file) + ".." + "fixtures"
+    let loader = FileSystemLoader(paths: [path])
+    environment = Environment(loader: loader)
+    template = ""
+    includedTemplate = ""
+  }
+
+  func testSyntaxError() throws {
+    template = Template(templateString: """
```
ilyapuchka: I don't see profit from using multi-line strings in this and next function 🤷‍♂️

Tests/StencilTests/EnvironmentSpec.swift
djbe: Ah ok.

> in a single line though
Problem is, and this in quite a few places, that we have absolutely gigantically long function calls/init/...

Thanks to swiftlint, we can enforce a hard limit on this, to improve readability. We can try keeping some things on one line, but here and there, we *will* have to split things up into multiple lines.

Tests/StencilTests/EnvironmentSpec.swift
djbe: It's to avoid having to escape `"` (quotes).

Tests/StencilTests/EnvironmentSpec.swift
ilyapuchka: yes, I agree, though having a function call that takes more than 5 lines probably is a sign that we should refactor it.

djbe: True, but there's not much of a choice here, we're passing along parameters to another function 🤷‍♂️

Tests/StencilTests/EnvironmentSpec.swift
```diff
@@ -1,342 +1,392 @@
...
+    environment = Environment(loader: ExampleLoader())
+    environment.extensions += [errorExtension]
+    template = ""
+  }
 
-      $0.it("can render a template from a string") {
-        let result = try environment.renderTemplate(string: "Hello World")
-        try expect(result) == "Hello World"
-      }
+  func testLoadingRendering() {
```
ilyapuchka: testLoading**And**Rendering, though I wonder if we better split it into two tests.

djbe: 👍

Tests/StencilTests/EnvironmentSpec.swift
ilyapuchka: I think escaping is fine... Don't we allow using single quotes though?

djbe: Sure, we could use `'`.

But you'll see in some other tests I've placed the template in multiline strings just to keep the formatting consistent between multiple tests. It's a nice visual queue of "hey, that's a template", or "hey, that's an expected result".

djbe:
Most of this PR is good to go, some of the feedback can't really be fixed without a more extensive refactor of the unit tests. 

I'm going to merge this for now, and further feedback/refactors can be done in other PRs. This one is an absolute hell to rebase after each PR 😱

PR #252, opened 2018-09-26
Fix lexer range calculation for tokens

djbe:
There's a bug where tokens without spaces such as `{{Test}}` are being parsed as `Tes`.

This PR fixes that issue, and adds a test to ensure this doesn't happen again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Stencil Changelog
 
+## Master
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


yonaskolb:
Was that a regression? Do we need a hotfix?

djbe:
This was a regression, yes. Unfortunately our tests didn't catch this before the 0.13 release.I've already made a 0.13.1 release, so no worries.

I meant to do a release PR, but I accidentally pushed on master and we're not allowed to rebase, so 🤷‍♂️

PR #254, opened 2018-09-26
Fix parsing token components with parenthesis without spaces

### LINKED ISSUE #253: Boolean expression: parenthesis need to be surrounded by spaces
Refs #165.

This is either a bug in the implementation, or in the documentation. The docs say:
`{% if (one or two) and three %}`
But that won't work. It'll throw a "dangling token" error. What does work is:
`{% if ( one or two ) and three %}`

The reason is that the current implementation is just built on top of the components array received from the token parser, which separates by spaces. @ilyapuchka thoughts?

ilyapuchka:
Resolves #253

### DIFF
--- a/Sources/Tokenizer.swift
+++ b/Sources/Tokenizer.swift
@@ -17,12 +17,12 @@ extension String {
           components[components.count-1] += word
         } else if specialCharacters.contains(word) {
           components[components.count-1] += word
-        } else if word != "(" && word.hasPrefix("(") {
-          components.append("(")
+        } else if word != "(" && word.hasPrefix("(") || word != ")" && word.hasPrefix(")") {
+          components.append(String(word.first!))
           appendWord(String(word.dropFirst()))
-        } else if word != ")" && word.hasSuffix(")") {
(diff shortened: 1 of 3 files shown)


djbe:
Just being pedantic here, but we want to make sure we cover all cases:

Does this handle a (badly written) expression like:
> (one and two )or three

AliSoftware:
Good point. If you're asking the question @djbe that means we need a test to cover it :wink:

ilyapuchka:
it won't handle ` )or` though it should, good catch

Sources/Tokenizer.swift
```diff
@@ -17,12 +17,12 @@ extension String {
           components[components.count-1] += word
         } else if specialCharacters.contains(word) {
           components[components.count-1] += word
-        } else if word != "(" && word.hasPrefix("(") {
-          components.append("(")
+        } else if word != "(" && word.hasPrefix("(") || word != ")" && word.hasPrefix(")") {
+          components.append(String(word.first!))
```
djbe: Please avoid force unwraps, especially in `Sources`.

You could:
- use `word[0]`
- unroll the if test into a separate one
- use `prefix(1)`
- ...

PR #256, opened 2018-09-26
Convert Token from enum to struct

### LINKED ISSUE #255: Convert Token from enum to struct with type property
Currently, `Token` is represented as enum which does not allow to store any properties in it, so we `components` property is computed, which affects performance as they can be parsed several times. We should instead convert it to `struct` and use a property of enum type like `type`.

ilyapuchka:
Resolves #255. Strangely enough, it makes parsing slightly slower, but I still think it's a good thing to do.
I didn't want to change too much code so I left static constructors instead of cases for interface to be the same.

### DIFF
--- a/Sources/Tokenizer.swift
+++ b/Sources/Tokenizer.swift
@@ -88,7 +88,7 @@ public class Token: Equatable {
   public let sourceMap: SourceMap
   
   /// Returns the underlying value as an array seperated by spaces
-  public lazy var components: [String] = self.contents.smartSplit()
+  public private(set) lazy var components: [String] = self.contents.smartSplit()
   
   init(contents: String, kind: Kind, sourceMap: SourceMap) {
     self.contents = contents
--- a/Sources/Tokenizer.swift
+++ b/Sources/Tokenizer.swift
@@ -71,47 +71,53 @@ public struct SourceMap: Equatable {
   }
 }
 
-public enum Token : Equatable {
+public class Token: Equatable {
+  public enum Kind: Equatable {
+    /// A token representing a piece of text.
+    case text
+    /// A token representing a variable.
+    case variable
+    /// A token representing a comment.
+    case comment
+    /// A token representing a template block.
+    case block
+  }
+  
+  public let contents: String
+  public let kind: Kind
+  public let sourceMap: SourceMap
+  
+  /// Returns the underlying value as an array seperated by spaces
+  public lazy var components: [String] = self.contents.smartSplit()
+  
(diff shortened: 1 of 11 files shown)


Sources/Tokenizer.swift
```diff
@@ -71,47 +71,53 @@ public struct SourceMap: Equatable {
...
+    /// A token representing a template block.
+    case block
+  }
+  
+  public let contents: String
+  public let kind: Kind
+  public let sourceMap: SourceMap
+  
+  /// Returns the underlying value as an array seperated by spaces
+  public lazy var components: [String] = self.contents.smartSplit()
```
AliSoftware: Shouldn't that be `private(set)`?

djbe: jep

djbe: We may also want to avoid splitting into components if the type is text?

Sources/Tokenizer.swift
```diff
@@ -71,47 +71,53 @@ public struct SourceMap: Equatable {
...
-  public var contents: String {
-    switch self {
-    case .block(let value, _),
-         .variable(let value, _),
-         .text(let value, _),
-         .comment(let value, _):
-      return value
-    }
+  
+  public static func == (lhs: Token, rhs: Token) -> Bool {
```
djbe: Don't we get equatable for free?

Sources/Tokenizer.swift
ilyapuchka: We can return just one item with the whole content, but I think this should be handled on a higher level and might as well not happening at all due to the logic flow

Sources/Tokenizer.swift
ilyapuchka: not for classes...

djbe:
Quick question: how did you measure a slowdown?

ilyapuchka:
in the performance test, it went from 250-something to 260-something

djbe:
The performance test is always a bit inconsistent though, you have to run it a few times to get an idea if it's a fluke, or an actual increase.

Anyway, 10ms increase isn't a problem in this case.

djbe:
What we're not noticing now, is a potential performance benefit in other parts of Stencil, as we only measure the lexer performance.

djbe:
Code looks GTM.

Might be worth adding documentation like I did in https://github.com/stencilproject/Stencil/pull/226/commits/cb4e51484660d17c92c0baa8e4c6f1e1b0f8e1a0, the current codebase can be quite complicated for developers to understand, and the docs should help with that.

Also, don't forget a changelog entry! 😉

PR #257, opened 2018-09-30
Shorthand syntax for filter tag

### LINKED ISSUE #244: Shorthand for `filter` tag
I found myself using `filter indent` quite often in one of the templates and it would be nice to be able to use

```
{% indent:4," ",true %}...{% endindent %}
```

instead of

```
{% filter indent:4," ",true %}...{% endfilter %}
```

This will not just make it shorter but will provide more context in the closing tag.

ilyapuchka:
Resolves #244

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ _None_
 
 ### Enhancements
 
-_None_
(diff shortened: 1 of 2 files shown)


ilyapuchka:
Is it something we may include or does not worth it and can be closed (see linked issue for context)

PR #258, opened 2018-10-01
Feat: Add support for Swift 4.0

Andrew-Lees11:
I am part of the Kitura team and we would like to update to the latest version of Stencil, however since we support Swift 4.0 and so are unable to use the latest version unless Stencil also supports Swift 4.0.

This pull requests adds support for Swift 4.0 by

- adding the auto-generated functions for equatable
- Using #if swift(>=4.1) to alternate between flatMap and compactMap
- Using #if swift(>=4.1) to alternate the requirement for where T.IndexDistance == Int with collections

Swift 4.0.3 builds have also been added to travis to check for future Swift 4.0 is supported in future builds.

These changes allow Stencil to compile and run on Swift 4.0 while still compiling with no build warnings for Swift 4.1

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.1
+// swift-tools-version:4.0
 import PackageDescription
 
 let package = Package(
--- a/Sources/Errors.swift
+++ b/Sources/Errors.swift
@@ -37,15 +37,8 @@ public struct TemplateSyntaxError : Error, Equatable, CustomStringConvertible {
   public init(_ description: String) {
     self.init(reason: description)
   }
-    
-  public static func ==(lhs: TemplateSyntaxError, rhs: TemplateSyntaxError) -> Bool {
-    return lhs.reason == rhs.reason &&
-    lhs.description == rhs.description &&
-    lhs.token == rhs.token &&
-    lhs.stackTrace == rhs.stackTrace &&
-    lhs.templateName == rhs.templateName &&
-    lhs.allTokens == rhs.allTokens
-  }
+  
+  
 }
 
 extension Error {
--- a/Sources/Errors.swift
+++ b/Sources/Errors.swift
@@ -37,6 +37,15 @@ public struct TemplateSyntaxError : Error, Equatable, CustomStringConvertible {
   public init(_ description: String) {
     self.init(reason: description)
   }
+    
+  public static func ==(lhs: TemplateSyntaxError, rhs: TemplateSyntaxError) -> Bool {
+    return lhs.reason == rhs.reason &&
+    lhs.description == rhs.description &&
+    lhs.token == rhs.token &&
+    lhs.stackTrace == rhs.stackTrace &&
+    lhs.templateName == rhs.templateName &&
+    lhs.allTokens == rhs.allTokens
+  }
 }
 
 extension Error {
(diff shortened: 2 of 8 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:4.1
+// swift-tools-version:4.0
```
ilyapuchka: does it still work when built with swift 4.1?...

Sources/Errors.swift
```diff
@@ -37,6 +37,15 @@ public struct TemplateSyntaxError : Error, Equatable, CustomStringConvertible {
...
     self.init(reason: description)
   }
+    
+  public static func ==(lhs: TemplateSyntaxError, rhs: TemplateSyntaxError) -> Bool {
+    return lhs.reason == rhs.reason &&
+    lhs.description == rhs.description &&
+    lhs.token == rhs.token &&
+    lhs.stackTrace == rhs.stackTrace &&
+    lhs.templateName == rhs.templateName &&
+    lhs.allTokens == rhs.allTokens
```
ilyapuchka: `allTokens` check is not needed as it is computed property. Also lines starting with 43 should be indented on one position.

Sources/IfTag.swift
```diff
@@ -118,7 +118,12 @@ final class IfExpressionParser {
   private init(components: ArraySlice<String>, tokenParser: TokenParser, token: Token) throws {
     var parsedComponents = Set<Int>()
     var bracketsBalance = 0
-    self.tokens = try zip(components.indices, components).compactMap { (index, component) in
+    #if swift(>=4.1)
```
ilyapuchka: maybe it would be better to add an extension that defines `compactMap` on collections for swift 4.1 rather than checking it in place? will be also easier to remove later.

Sources/Errors.swift
djbe: I'd also wrap this whole `==` operator in a `#if swift` check.

Sources/Variable.swift
```diff
@@ -116,9 +116,32 @@ public struct Variable : Equatable, Resolvable {
 
     return normalize(current)
   }
+    
+  public static func ==(lhs: Variable, rhs: Variable) -> Bool {
```
djbe: Same here, wrap in `#if swift`.

djbe:
I agree that we should define `compactMap` for older swift versions, and maybe we should do this in a `_SwiftSupport.swift` file, where we can collect all these backwards compat. fixes.

The `==` can also be implemented there.

Package.swift
Andrew-Lees11: This will still work. The swift-tools-version specifies the minimum version of Swift required for the project but higher versions are allowed.

Andrew-Lees11:
Hey
Thank you for the quick review. As suggested I have moved all the changes into a single file. This extends Collection to so future compact maps and index should work for Swift 4.0 as well.

Sources/_SwiftSupport.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+#if swift(>=4.1)
```
djbe: I misread that at first and thought you made a typo, until I saw the `else` clause.
Maybe make this `#if !swift(>=4.1)`?

djbe:
Note for those wondering: the file should be named `_SwiftSupport` so that it's the first to compile, otherwise Swift doesn't always find the extensions.

Issue #120, opened 2017-05-31
Remove Dependency of PathKit for External Programs


siilime:
There's no logical reason that a consuming program would need PathKit. An overloaded `init` could be included to support a plain `String` representation of a path that is internally converted to a PathKit `Path`.

This would remove an unnecessary third-party dependency.

```
public init(from paths: [String]) {
    ...
}
```

kylef:
> An overloaded init could be included to support a plain String representation of a path that is internally converted to a PathKit Path.

- Path is string literal convertible. You can pass literal strings to the same method. As show in the documentation (https://stencil.fuller.li/en/latest/api.html#filesystemloader)

    ```swift
    FileSystemLoader(paths: ["./templates"])
    ```

- Path adds an additional type hint so it is clear what the intended value is.

> This would remove an unnecessary third-party dependency

PathKit is not a third-party dependency. It is first party, I made both Stencil and PathKit.

basememara:
I agree in removing PathKit dependency. The location of the template has nothing to do with the actual template parsing engine which is the core purpose. First-party, third-party, etc it's still dependency and removing would make `Stencil` much more portable and lighter.

djbe:
- PathKit is a very small dependency, so "much lighter" isn't an argument.
- PathKit works on Linux, so it is portable.
- Stencil needs to load files, so it needs filesystem access, in a platform independent way. Foundation is a bit lacking in this sense, especially with path transformations, etc...

basememara:
How about merging it in then if it's so important to the core functionality. Juggling dependencies and waiting for both to be in sync as far as versioning and compatibility is a bit clumsy.

djbe:
How are we juggling dependencies? Stencil defines which PathKit version it needs, and is compatible with. Your package manager (CocoaPods/SPM) will install the correct version accordingly.

PathKit is used by other projects, it makes no sense to directly merge it into Stencil, as that would mean we'd need to maintain 2 codebases.

basememara:
I see your points. It's just that having a library depending on another library creates more moving parts. A self-contained library would be easier to manage and consume. Even if just extracting the piece of `PathKit` that `Stencil` needs would be great.

djbe:
Like I mentioned before, you don't need to manage, nor consume the library yourself. You tell CocoaPods or SPM you need Stencil 0.13, and they get it for you + any dependencies.

basememara:
[...]

PR #259, opened 2018-10-01
Decouple from PathKit

basememara:
Remove FileSystemLoader in case not needed for projects and break PathKit coupling. Further discussion in #120.

### DIFF
--- a/Sources/Include.swift
+++ b/Sources/Include.swift
@@ -1,4 +1,3 @@
-import PathKit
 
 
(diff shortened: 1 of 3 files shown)


djbe:
The file system loading is necessary for the `include` tag, as well as the template inheritance.

Look, I honestly don't think we're going to merge this. You'd break all the other frameworks and apps already using Stencil, just because you don't want to add one line to your Carthage file?

PR #261, opened 2018-12-12
Stencil is used by Genesis

tomj:
I guess in a perfect world there'd be some bot which searches GH for deps and lists them for a project but right now this should suffice 😂

Thanks for a great project in Stencil! 🎉


yonaskolb:
Thanks @tomj!

PR #267, opened 2019-01-21
always evaluate block.super even if it is not used

### LINKED ISSUE #266: Blocks and if don't seem to mix
Both of these don't do what I expected them to do (taken from stencil that extends another one):

```
{% block title %}
    {% if sort == "new" %}
        {{ block.super }} - Nieuwste spellen
    {% elif sort == "upcoming" %}
        {{ block.super }} - Binnenkort op de agenda
    {% elif sort == "near-me" %}
        {{ block.super }} - In mijn buurt
    {% endif %}
{% endblock %}
```

```
{% if sort == "new" %}
    {% block title %}{{ block.super }} - Nieuwste spellen{% endblock %}
{% elif sort == "upcoming" %}
    {% block title %}{{ block.super }} - Binnenkort op de agenda{% endblock %}
{% elif sort == "near-me" %}
    {% block title %}{{ block.super }} - In mijn buurt{% endblock %}
{% endif %}
```

In the former, only "- Nieuwste spellen" (for example) shows and `{{ block.super }}` evaluates to nothing.
In the latter, the reverse happens.

What does work is:

```
{% block title %}
    {{ block.super }} -
    {% if sort == "new" %}
         Nieuwste spellen
    {% elif sort == "upcoming" %}
        Binnenkort op de agenda
    {% elif sort == "near-me" %}
        In mijn buurt
    {% endif %}
{% endblock %}
```

ilyapuchka:
Resolves #266 

The issue is caused by the fact that we were trying to avoid rendering `block.super` when it is not used in a child template by searching for such variable inside `block` node. The problem is that if inside this node there are other nodes, i.e. `{% if %}` then we are not searching inside them. We could potentially search for it recursively, but this can lead to unneeded code and runtime complexity. Instead we can always render it when rendering `extend` node.

This will come at a cost of possibly unneeded rendering of the block (which still may be cheeper than going through nodes tree searching for it) and any errors in the base block being reported for `{% extends %}` node (see changed test case) instead of `{{ block.super }}`. 

Searching for `block.super` would not work reliably for error reporting any way because we are searching only for first occasion of `block.super` node, where at runtime actual rendering may be happening in another place (i.e. in a different branch of `if` node).

Alternatively we could render the content of `block.super` lazily by storing closure instead of rendered value in the context, but quick change like that resulted in infinite recursion and probably requires more changes to make it work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,10 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Fixed using `{{ block.super }}` inside nodes other than `block`.  
+  [Ilya Puchka](https://github.com/ilyapuchka)
+  [#266](https://github.com/stencilproject/Stencil/issues/266)
+  [#267](https://github.com/stencilproject/Stencil/pull/267)
 
 ### Internal Changes
 
--- a/Sources/Inheritence.swift
+++ b/Sources/Inheritence.swift
@@ -136,7 +136,11 @@ class BlockNode: NodeType {
 
   func render(_ context: Context) throws -> String {
     if let blockContext = context[BlockContext.contextKey] as? BlockContext, let child = blockContext.pop(name) {
-      let childContext = try self.childContext(child, blockContext: blockContext, context: context)
+      let childContext: [String: Any] = [
+        BlockContext.contextKey: blockContext,
+        "block": ["super": try self.render(context)]
+      ]
+
       // render extension node
       do {
         return try context.push(dictionary: childContext) {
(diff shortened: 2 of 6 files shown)


ilyapuchka:
@djbe I don't think I have it

SwiftGen-Eve:

---

</p>

djbe (approved):
I played around with lazy evaluation (see #324), but kept hitting generation order issues. I presume it's because the `BlockContext` is also modified during renders, causing issues with lazy-rendering.

These changes are good as is (after rebase). Yes there'll be a performance impact, but it'll fix an error, so it's worth it.

PR #268, opened 2019-01-29
Support swift 5.0

kawoou:
* Warning: 'index(where:)' is deprecated: renamed to 'firstIndex(where:)'
 * Add Package@swift-5.swift
 * Update podspec swift version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,17 +9,24 @@ matrix:
     - os: osx
       osx_image: xcode10
       env: SWIFT_VERSION=4.2
+    - os: osx
+      osx_image: xcode10.2
+      env: SWIFT_VERSION=5.0
     - os: linux
       env: SWIFT_VERSION=4.0.3
     - os: linux
       env: SWIFT_VERSION=4.1
     - os: linux
       env: SWIFT_VERSION=4.2
+    - os: linux
+      env: SWIFT_VERSION=4.2.3
+    - os: linux
+      env: SWIFT_VERSION=5.0-DEVELOPMENT-SNAPSHOT-2019-03-10-a
 language: generic
 sudo: required
 dist: trusty
 install:
-  - eval "$(curl -sL https://swiftenv.fuller.li/install.sh)"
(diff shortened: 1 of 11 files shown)


djbe:
That was quick! 😆 

Are there other things we can potentially migrate? There are new `String` protocols, and other things (see https://www.whatsnewinswift.com).

Instead of peppering the code with `#if swift` checks, we add backports for new functions. See the `_SwiftSupport.swift` file:
https://github.com/stencilproject/Stencil/blob/master/Sources/_SwiftSupport.swift

soffes:
Any word on this PR? I'd love to update to Swift 5 :)

kawoou:
@djbe I want to support #"..."# syntax for readability.
However, you must also support versions below Swift 5.
If I do, think the code will be less readable.

ex)
```swift
#if swift(<5.0)
self.format = format ?? Variable("\"yyyy-MM-dd 'at' HH:mm\"")
#else
self.format = format ?? Variable(#""yyyy-MM-dd 'at' HH:mm""#)
#endif
```

djbe:
Is there more info about that SIL exception? Is it new in Swift 5? Radar?

djbe:
Yeah, we won't be able to use raw strings until we drop 4.2 support 🤷‍♂️

kawoou:
@djbe Compile-time keywords(like #file) are expected to cause problems due to type inference errors during static dispatch.

**Before:** Throw SIL Exception (https://travis-ci.org/stencilproject/Stencil/jobs/508764850)
```swift
let path = Path(#file) + ".."  + "fixtures" + "huge.html"
```
**After:**
```swift
let basePath: String = #file //< Specify type
let path = Path(basePath) + ".."  + "fixtures" + "huge.html"
```

djbe:
Would this work then?
```swift
let path = Path(#file as String) + ".."  + "fixtures" + "huge.html"
```

keith:
This changed fixed PathKit https://github.com/kylef/PathKit/pull/69/files#diff-663087ac5b85edfc8e4ff1dd83b915abL12

soffes:
Awesome! Thanks so much for your effort on this!

Mind tagging a release with Swift 5 support?

PR #277, opened 2019-05-07
Replaced deprecated Swift function

philippec-ls:
index(of:) is deprecated in Swift 5, replaced with firstIndex(of:).

This change is compatible with Swift 4.2.


djbe:
Thank you for this PR.

I'm going to close this PR though, as there's already one for Swift 5 support that is a bit more extensive, and keeps compatibility with Swift 4.

philippec-ls:
Thanks for the update. I realized later there was another PR for Swift 5, so I apologize for wasting time but I hope the other PR will be merged soon :-)

PR #281, opened 2019-07-17
Allow context to be initialised with Any

adam-fowler:
Given Mirror is being used to serialize objects that are members of the Context dictionary, it only seems logical we shouldn't be limited to initializing a Context with a dictionary. This PR adds the ability to initialize a context with Any

Initialize Context with Any by converting to dictionary
Allow Any to be pushed onto a Context in a similar way
Change Environment functions to use these new functions

### DIFF
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -14,6 +14,10 @@ public class Context {
     self.environment = environment ?? Environment()
   }
 
+  convenience init(object: Any? = nil, environment: Environment? = nil) {
+    self.init(dictionary: Context.dictionaryFromAny(object: object ?? [:]), environment: environment)
+  }
+    
   public subscript(key: String) -> Any? {
     /// Retrieves a variable's value, starting at the current context and going upwards
     get {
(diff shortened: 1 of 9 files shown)


adam-fowler:
This will fix https://github.com/stencilproject/Stencil/issues/223#issuecomment-411833557

Tests/StencilTests/TemplateSpec.swift
```diff
@@ -6,13 +6,13 @@ final class TemplateTests: XCTestCase {
   func testTemplate() {
     it("can render a template from a string") {
       let template = Template(templateString: "Hello World")
-      let result = try template.render([ "name": "Kyle" ])
+      let result = try template.render(object:[ "name": "Kyle" ])
```
ilyapuchka:
```suggestion
      let result = try template.render(object: ["name": "Kyle"])
```

Tests/StencilTests/TemplateSpec.swift
```diff
@@ -6,13 +6,13 @@ final class TemplateTests: XCTestCase {
...
       try expect(result) == "Hello World"
     }
 
     it("can render a template from a string literal") {
         let template: Template = "Hello World"
-        let result = try template.render([ "name": "Kyle" ])
+        let result = try template.render(object:[ "name": "Kyle" ])
```
ilyapuchka:
```suggestion
        let result = try template.render(object: ["name": "Kyle"])
```

ilyapuchka (approved):

Sources/Context.swift
```diff
@@ -14,6 +14,10 @@ public class Context {
     self.environment = environment ?? Environment()
   }
 
+  convenience init(object: Any? = nil, environment: Environment? = nil) {
```
ilyapuchka:
```suggestion
  convenience public init(object: Any? = nil, environment: Environment? = nil) {
```

adam-fowler:
@ilyapuchka Is this getting merged in or do I need another approval?

Sources/Context.swift
```diff
@@ -45,13 +49,29 @@ public class Context {
     return dictionaries.popLast()
   }
 
+  /// return a dictionary describing an object
+  static fileprivate func dictionaryFromAny(object: Any) -> [String: Any] {
+    if let dictionary = object as? [String: Any] {
+      return dictionary
+    } else {
+      let dictionary = Mirror(reflecting: object).asDictionary()
+      return dictionary
```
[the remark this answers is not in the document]
djbe: Simplify this into 1 line

djbe (changes requested):
Can we possibly avoid the `Mirror(...).asDictionary()` bit? Stencil already has quite extensive reflection capabilities for multiple types (when rendering a template).

adam-fowler:
> Can we possibly avoid the `Mirror(...).asDictionary()` bit? Stencil already has quite extensive reflection capabilities for multiple types (when rendering a template).

I'll have to have a good look at the code again. It was a wee while ago when I wrote this

adam-fowler:
> Can we possibly avoid the `Mirror(...).asDictionary()` bit? Stencil already has quite extensive reflection capabilities for multiple types (when rendering a template).

@djbe I had a look through the code again. None of the reflection code seems accessible at this point. Is there function that takes an object and converts it to a `[String: Any]` somewhere else?

PR #286, opened 2019-10-27
Fix for incorrect tokenization due to index difference of Unicode character/scalar

### LINKED ISSUE #276: Unknown template tag 'else'
Reraising an issue from Sourcery repository: https://github.com/krzysztofzablocki/Sourcery/issues/735 

With the very latest release of Sourcery, my templates (that were 100% working) are now completely broken.
Can you please help me with understanding what is incorrect in my template?

andreasley:
Probably fixes #276 (does for my case).

`Scanner` iterates unicode scalars but would previously split strings by character index. Since multiple unicode scalars might be combined into a single character, this could lead to incorrect tokenization and therefore unknown tags.
This patch changes `Scanner` to use the indexes of the respective `UnicodeScalarView`.

All tests pass and performance is basically unchanged.

Bigger picture: I wonder if `Unicode.Scalar` should be used at all in `Lexer`/`Scanner`, as splitting is conceptually always done by character.


Tests/StencilTests/LexerSpec.swift
```diff
@@ -126,6 +126,17 @@ final class LexerTests: XCTestCase {
     }
   }
 
+  func testCombiningDiaeresis() throws {
+    let path = Path(#file as String) + ".."  + "fixtures" + "combining-diaeresis.html"
```
ilyapuchka: can we maybe just inline the content of the file here? fixtures are really used to test logic specific to template inheritance and such stuff that requires files

andreasley: I wasn't sure how special Unicode symbols might affect editors and build tools, but at least Xcode 11.2.1 and its toolchain seem to be fine with it. :)

Issue #248, opened 2018-09-23
Revisit behaviour of undefined values


ilyapuchka:
Current behavior of `Variable.resolve` when it encounters undefined values is IMO inconsistent with other kinds of errors, like an unknown filter, which results in an exception. Accessing undefined value, i.e. `blog.uri` instead of `blog.url` results in the empty output instead of exception or JavaScript-like output "undefined" which can be confusing for users.

Before #245 when resolving properties of collections unknown keys were simply skipped, resulting in `contacts.something.0` render the same way as `contacts.0` instead of raising an exception or returning `nil`. With #245 resolving arrays and string properties was refactored so it will now return `nil`. But the issue is still present for dictionaries.

Expected behavior:

- accessing undefined property should result in an exception by default, we can add configuration to allow ignoring such errors and printing them out as a warning or completely ignoring them and render empty string, "nil" or anything else
- rendering `nil` values can still result in an empty string or can be configured to render "undefined" or anything else per user choice, i.e. "nil"/"null"
- resolving unknown properties on any collection type should behave the same way as resolving unknown properties of any other type
- accessing collection elements out of bounds should result in exception by default but can be configurable to warn in console or completely ignore this and render empty string, "nil" or anything else

ilyapuchka:
Yes, this should be configurable to some degree. Though what do you mean "don't know if a property exists or not"? We can add `is defined` check to test for that, i.e. when resolving dictionary keys (to check if key contains `nil` or is not present), but IMO as we are talking about Swift environment there shouldn't be such thing as "undefined" on any other type. Also should accessing out of bounds index in collection result in nil or exception? I'd say exception, but can be also be configurable.

djbe:
Nah I mean for example:
- Stencil is configured to the "throw exceptions" mode
- I write a template, but at some point, some property (or variable, ...) may or may not exist
- As a template writer, before I trigger an exception, I'd like to test if the property/variable/... exists

ilyapuchka:
By "may not exist" you mean "value of the property is `nil` at runtime" or "there is no such property on the type at all"? 

In the first case, I think it should not through any exception by default and there should be a way to check for that at runtime, like `!= nil` in Swift. But we can have a "strict" mode that will treat accessing any `nil` values as errors. 

The second case should by default result in the exception, but this can be softened via configuration.

djbe:
I think there may be multiple levels? And a `!= nil` check is needed.

[...]

PR #289, opened 2019-12-26
Added the option to throw when variable not found in context

acecilia:
Added the option to throw when a variable used in a template could not be resolved, which most probably means it was not defined in the context


acecilia:
@kylef @djbe @AliSoftware Could you have a look?

Sources/Variable.swift
```diff
@@ -30,10 +30,16 @@ class FilterExpression: Resolvable {
   func resolve(_ context: Context) throws -> Any? {
     let result = try variable.resolve(context)
 
-    return try filters.reduce(result) { value, filter in
+    let filteredResult = try filters.reduce(result) { value, filter in
       let arguments = try filter.1.map { try $0.resolve(context) }
       return try filter.0.invoke(value: value, arguments: arguments, context: context)
     }
+
+    if filteredResult == nil, context.environment.throwOnUnresolvedVariable {
```
ilyapuchka: I don't fill like this is a correct place for this check. nil can be a result of filter operation itself. Also it does not cover usage of undefined variables outside of filter. 
What we need to do is to check for undefined variable in `Variable.resolve` method when we do `if current == nil { return nil }` and to throw instead of returning nil.

acecilia: Thanks for the feedback! :) I will break it down into some points:
a) "nil can be a result of filter operation itself": yes. This is the reason why I implemented this feature: the main idea behind it is that getting a nil in a template indicates a bug (either because the variable is not defined or because of a filter failing to find a matching result), and instead of silently not printing anything, I would prefer to throw.
b) "Also it does not cover usage of undefined variables outside of filter": I do not fully understand what you mean, seems to be working correctly as per the test `testUnresolvedVariable` I added. I am not very familiar with the Stencil codebase, can you give me an example of template where this will not work?
c) "What we need to do is to check for undefined variable in Variable.resolve method": I also tried that, but then the `default` filter will be useless, as we will be throwing even when a default value is provided.

acecilia: Maybe instead of `throwOnUnresolvedVariable` I should call the option `throwOnUnresolvedExpression`?

acecilia: @ilyapuchka could you have a second look at this?

acecilia:
Related to https://github.com/stencilproject/Stencil/issues/248

djbe:
What'd be nice is if we could have a `ResolveMode` enum (bad at names), as mentioned in #248:
- Unchecked: accessing `nil` is okay, non-existing property is okay (returns `nil`)
- Strict: accessing `nil` is okay, non-existing property throws
- StrictNoNil: accessing `nil` throws, non-existing property throws

Although it may even be an OptionSet, so that users can configure these things individually (nil access, non-existing access). That way, at a later point we may define extra behaviours such as KVO selectors, etc...

AliSoftware:
For the name of such an enum I propose `PropertyFallbackBehavior`

acecilia:
@djbe Can you put some usecase examples? The second one you mention I do not see the usecase:

* Unchecked: thats the current way.
* Strict: I do not see the usecase for this: under which scenario will you be accessing `nil` if a property returned a value correctly? And in that case, why would you like it not throwing?
* StrictNoNil: the implementation of this PR, where it throws if the final value after applying all the filters is nil.

acecilia:
Also, in the `StrictNoNil` (this PR), if I expect something may be nil and I do not want to throw, what I am doing is:

```
{{ name|default::"" }}
```

djbe:
Right the `strict` mode I propose above would avoid having to add `|default` everywhere, but at least throw errors for unknown variables (typos, etc...)

acecilia:
I can look into it, but, is there any possibility of getting it merge? Just asking because I see many unmerged PRs in this repo.

Also, are you okey with the current implementation for `StrictNoNil` in this PR? Because @ilyapuchka didnt like it so much.

Thanks for answering :)

PR #294, opened 2020-05-26
Drop Swift 4 support

djbe:
As Swift 4 is really old (by now), and our dependencies have dropped support for it (PathKit), we may as well drop support for Swift 4 ourselves.


Package.swift
```diff
@@ -18,5 +18,6 @@ let package = Package(
       "Stencil",
       "Spectre"
     ])
-  ]
+  ],
+  swiftLanguageVersions: [.v4, .v4_2]
```
yonaskolb, line 22: Should `.v4` be in here?

PR #297, opened 2020-08-02
Update PathKit in Package.swift

fortmarek:
`Package@swift-5.swift` has a newer version of `PathKit` than `Package.swift` where `Package.swift` contains an outdated reference to `PathKit` `0.9.0`.

I think it'd make sense to update it 🙂


AliSoftware:
Thanks for the PR!

PR #305, opened 2021-01-18
Fix build warning

danpalmer:
`components` is not mutated, so it can be a `let`. This fixes the build warning that otherwise shows up in build logs.


danpalmer:
CI failure appears unrelated - a network timeout. Unfortunately I can't re-run.

kylef (approved):

PR #306, opened 2021-04-09
Fix for crashing range indexes when variable length is 1

lkuczborski:
This PR fixes a crash when variable length equals 1 - one single quote (') or one double quote("):
`Fatal error: Can't form Range with upperBound < lowerBound: file Swift/x86_64-apple-macos.swiftinterface, line 14814`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ _None_
 - Throw syntax error on empty variable tags (`{{ }}`) instead `fatalError`.  
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#263](https://github.com/stencilproject/Stencil/pull/263)
+- Fix for crashing range indexes when variable length is 1.  
+  [Łukasz Kuczborski](https://github.com/lkuczborski)
+  [#306](https://github.com/stencilproject/Stencil/pull/306)
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ _None_
 
 ### Bug Fixes
 
-_None_
+- Fix for crashing range indexes when variable length is 1.  
+  [Łukasz Kuczborski](https://github.com/lkuczborski)
+  [#306](https://github.com/stencilproject/Stencil/pull/306)
 
 ### Internal Changes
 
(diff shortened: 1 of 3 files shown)


Sources/Variable.swift
```diff
@@ -50,6 +50,7 @@ public struct Variable: Equatable, Resolvable {
   public func resolve(_ context: Context) throws -> Any? {
     if (variable.hasPrefix("'") && variable.hasSuffix("'")) || (variable.hasPrefix("\"") && variable.hasSuffix("\"")) {
       // String literal
+      guard variable.count > 1 else { return "" }
```
djbe, line 53: Move this logic to the if test:
```swift
if variable.count > 1 && (...) {
```

djbe (changes requested):
Thank you for the PR!

Besides the small tweak, please add a changelog entry in the same format as the other entries 🙏 (easiest is to copy an existing entry, the spaces are important for formatting)

Sources/Variable.swift
lkuczborski: @djbe Done! :)

lkuczborski: @djbe Fixed logic and tests and added changelog entry. Build passed 🟢

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ _None_
 - Throw syntax error on empty variable tags (`{{ }}`) instead `fatalError`.  
   [Ilya Puchka](https://github.com/ilyapuchka)
   [#263](https://github.com/stencilproject/Stencil/pull/263)
+- Fix for crashing range indexes when variable length is 1.
```
djbe, line 52: That's not the right place 😄 
Put it in the `Master` section under bug fixes.

djbe (changes requested):

krzysztofzablocki:
great, this fixes a crash that occurs in the middle of template writing in Sourcery

CHANGELOG.md
lkuczborski: Fixed.

krzysztofzablocki (approved):

krzysztofzablocki:
@djbe good to merge? I'll need to grab this for next Sourcery Pro release

djbe (approved):

PR #313, opened 2021-09-22
Update Package.swift

neha275:
update Spectre library version


urosMihailovic (approved):

urosMihailovic (approved):

urosMihailovic (commented):
Maybe this would also be a good time to update PathKit dependency to 1.0.1 since it also solves Xcode 13 related issues.

PR #314, opened 2021-09-23
Fix Xcode 13 compilation issue by updating PathKit to 1.0.1

astromonkee:


marcosgriselli:
Thanks for merging this @djbe! Will there be an automatic release or it needs to be created manually?

PR #324, opened 2022-07-28
Support lazy context data

djbe:
This adds support for providing lazily calculated context data. It provides 3 ways of doing this:
- Plain lazy data, when you don't need the context: `LazyValueWrapper(someMagicStuff())`
- Lazy data using context: `LazyValueWrapper { context in … }`
- Lazy data using a copy of the context (for example in custom tags): `LazyValueWrapper(copying: context) { context in … }`

An example where this is useful is in SwiftGen, where we want to avoid doing some heavier metadata parsing that most users won't need. We made our own implementation, which can be replaced with this PR.
https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Stencil/ContextHelpers.swift

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@
   [David Jennes](https://github.com/djbe)
   [#164](https://github.com/stencilproject/Stencil/pull/164)
   [#325](https://github.com/stencilproject/Stencil/pull/325)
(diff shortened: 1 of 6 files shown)


SwiftGen-Eve:

---

</p>

PR #335, opened 2023-02-01
Add 'unique' filter

DavidBertet:
Returns a list of unique items from the given value.

    {{ value|unique }}

The unique items are yielded in the same order as their first occurrence in the iterable passed to the filter.


DavidBertet:
@djbe is this repo still maintained? Is there a chance to get that merge at some point?

Thanks!

PR #341, opened 2023-06-25
Stencil: use unstable features for portability

compnerd:
Rather than exhaustively enumerating via OS use the unstable feature to detect if ObjC interop is enabled.  This allows building on Windows.

Unfortunately, the entire test suite does not pass on Windows, with path issues due to PathKit's handling of paths.

### DIFF
--- a/Sources/Stencil/Variable.swift
+++ b/Sources/Stencil/Variable.swift
@@ -111,12 +111,12 @@ public struct Variable: Equatable, Resolvable {
     } else if let string = context as? String {
       return resolve(bit: bit, collection: string)
     } else if let object = context as? NSObject {  // NSKeyValueCoding
-      #if os(Linux)
-        return nil
-      #else
+      #if _runtime(_ObjC)
         if object.responds(to: Selector(bit)) {
           return object.value(forKey: bit)
         }
+      #else
+        return nil
       #endif
     } else if let value = context as? DynamicMemberLookup {
       return value[dynamicMember: bit]


compnerd:
This lays the groundwork for enabling Windows support.

art-divin:
Hello @compnerd ,

thank you very much for your MR, it has given me a clue when I was investigating an issue with variable parsing under Linux for my Sourcery port: https://github.com/krzysztofzablocki/Sourcery/pull/1188.

I have created #342 which changes the order of checks, and makes it possible to have a subclass from `NSObject` till, but also to conform to `DynamicMemberLookup` protocol, thus giving manual control for KVC-emulation on non-Darwin platforms, maybe that would be of interest to you.

djbe:
@compnerd can I assume that @art-divin's PR supersedes this one? I.e. is `#if canImport(ObjectiveC)` equivalent to `#if _runtime(_ObjC)`? 

If it's not the case, please feel free to reopen this PR, or adding your comment in #342.

compnerd:
They are not equivalent - `canImport` does not speak to the runtime, it it simply indicates that a module with that name is available  I can simply provide a module named ObjectiveC and it will incorrectly assume to be true and fail to build.

djbe:
I can't seem to find any mentions towards using `_runtime(ObjC)`…

Only found some discussions about windows+swift+objc, that referred to https://github.com/apple/swift/issues/56739 which in turn referred to [hasAttribute(…)](https://github.com/apple/swift-evolution/blob/main/proposals/0367-conditional-attributes.md). But there's no mention of what attributes are available 😞

compnerd:
Yeah, `_runtime(ObjC)` is pretty rarely used, I think that there may be a few uses of it in Foundation.  But documentation is unlikely to be found - it is an internal (and thus unstable) feature.  The Windows + Swift + ObjC is definitely one way that the use of `canImport(ObjectiveC)` would return true, but in that case, I expect the KVC to be available.

It is more that there is no guarantee that I do not simply create a module with the name ObjectiveC.  Something like adding a module.modulemap with the following should be sufficient to evaluate the branch to true but not provide KVC:

```
module ObjectiveC [system] { }
```

djbe:
I get that it's theoretically possible to define a module called ObjectiveC, but is that something that people do on Swift in Windows (or other non-Darwin platforms)? Sounds a bit like a footgun.

I don't mind using the `_runtime` if there's a realistic case for it, otherwise we might just stick with `canImport` 🤷

compnerd:
Sure, sticking with that for now is probably okay.  I tend to try to be more proactively careful about the possibility.

PR #342, opened 2023-08-20
Prefer DynamicMemberLookup over KVC

art-divin:
## Context

On non-Darwin platforms KVC is not supported, thus check in `Variable.swift:115` is returning `nil`.

While this technically cannot be solved, it would be better  to first attempt to get value via `DynamicMemberLookup` protocol conformance.

This change is required for Sourcery linux support: https://github.com/krzysztofzablocki/Sourcery/pull/1188

### DIFF
--- a/Sources/Stencil/Variable.swift
+++ b/Sources/Stencil/Variable.swift
@@ -110,16 +110,16 @@ public struct Variable: Equatable, Resolvable {
       return resolve(bit: bit, collection: array)
     } else if let string = context as? String {
       return resolve(bit: bit, collection: string)
+    } else if let value = context as? DynamicMemberLookup {
+      return value[dynamicMember: bit]
     } else if let object = context as? NSObject {  // NSKeyValueCoding
-      #if os(Linux)
+      #if canImport(ObjectiveC)
         return nil
       #else
         if object.responds(to: Selector(bit)) {
           return object.value(forKey: bit)
         }
       #endif
-    } else if let value = context as? DynamicMemberLookup {
-      return value[dynamicMember: bit]
     } else if let value = context {
       return Mirror(reflecting: value).getValue(for: bit)
     }
--- a/Sources/Stencil/Variable.swift
+++ b/Sources/Stencil/Variable.swift
@@ -113,7 +113,7 @@ public struct Variable: Equatable, Resolvable {
     } else if let value = context as? DynamicMemberLookup {
       return value[dynamicMember: bit]
     } else if let object = context as? NSObject {  // NSKeyValueCoding
-      #if canImport(ObjectiveC)
+      #if !canImport(ObjectiveC)
         return nil
       #else
         if object.responds(to: Selector(bit)) {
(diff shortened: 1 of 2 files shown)


art-divin:
👋🏻  Hello @djbe , 

I am finishing Linux support for Sourcery.

At the moment this line of code is a blocker, would it be possible to merge this MR anytime soon?

I can see that this repo is unmaintained. I have resources to setup CI here and maybe I'll need to merge some more MRs before Sourcery Linux port would be finished.

I'd love to be collaborator here and support the project 🙏🏻

Sources/Stencil/Variable.swift
```diff
@@ -110,16 +110,16 @@ public struct Variable: Equatable, Resolvable {
       return resolve(bit: bit, collection: array)
     } else if let string = context as? String {
       return resolve(bit: bit, collection: string)
+    } else if let value = context as? DynamicMemberLookup {
+      return value[dynamicMember: bit]
     } else if let object = context as? NSObject {  // NSKeyValueCoding
-      #if os(Linux)
+      #if !canImport(ObjectiveC)
```
djbe, line 116: Tiny nitpick: I'd switch these around, so have the "positive" case first, as we do something there. I know that you didn't change the order, it's just something we can correct now.

djbe (changes requested):
Unfortunately the danger step doesn't work from other repos… Anyway, you're missing a changelog entry.

📝 We use the following format for CHANGELOG entries:
```
* Prefer DynamicMemberLookup over KVC.  
  [##342](https://github.com/stencilproject/Stencil/pull/342)
  [@art-divin](https://github.com/art-divin)
```
:bulb: Don't forget to end the line describing your changes by a period and two spaces.

Sources/Stencil/Variable.swift
djbe: Extra note: @compnerd mentioned that `canImport` isn't a sufficient check, so we'll probably need that `_runtime` check, or hopefully something more "official" than that.

art-divin: Hello @djbe ,

thank you for your comment.

I would like to address your points:

> I'd switch these around, so have the "positive" case first

Totally making sense, I missed this, and will apply this change right afterwards.

> Extra note: @compnerd mentioned that canImport isn't a sufficient check, so we'll probably need that _runtime check, or hopefully something more "official" than that.

As far as I have tried, using the (deprecated) `swift test --generate-linuxmain` generates `LinuxMain.swift` and a list of `XCTestManifests.swift` files, where this check is included:

```swift
#if !canImport(ObjectiveC)
import XCTest
```

So I would rather stick with something SPM engineering decided to go with, rather than being extra safe. Not only that, it is undefined if `_runtime` will still work in Swift 6.0 or future versions since it is internal.

djbe: I know, not a fan of using internal Swift functionality. We discussed it a bit further in there, and the current conclusion is that `canImport` is sufficient.

djbe (approved):

PR #343, opened 2023-08-27
Windows Support

compnerd:
In order to support Windows, PathKit must be removed.  Porting PathKit turned out to be more difficult than anticipated and it was easier to remove that in favour of Foundation.

### DIFF
--- a/Sources/Stencil/Loader.swift
+++ b/Sources/Stencil/Loader.swift
@@ -5,7 +5,6 @@
 //
 
 import Foundation
-import PathKit
 
 /// Type used for loading a template
 public protocol Loader {
@@ -34,15 +33,17 @@ extension Loader {
 
 // A class for loading a template from disk
 public class FileSystemLoader: Loader, CustomStringConvertible {
-  public let paths: [Path]
+  public let paths: [String]
 
-  public init(paths: [Path]) {
-    self.paths = paths
+  public init(paths: [URL]) {
+    self.paths = paths.map {
+      $0.withUnsafeFileSystemRepresentation { String(cString: $0!) }
+    }
   }
 
   public init(bundle: [Bundle]) {
-    self.paths = bundle.map { bundle in
-      Path(bundle.bundlePath)
+    self.paths = bundle.map {
+      URL(fileURLWithPath: $0.bundlePath).withUnsafeFileSystemRepresentation { String(cString: $0!) }
     }
   }
 
@@ -51,27 +52,19 @@ public class FileSystemLoader: Loader, CustomStringConvertible {
   }
 
   public func loadTemplate(name: String, environment: Environment) throws -> Template {
-    for path in paths {
-      let templatePath = try path.safeJoin(path: Path(name))
-
-      if !templatePath.exists {
-        continue
-      }
-
-      let content: String = try templatePath.read()
-      return environment.templateClass.init(templateString: content, environment: environment, name: name)
-    }
-
-    throw TemplateDoesNotExist(templateNames: [name], loader: self)
+    return try loadTemplate(names: [name], environment: environment)
   }
 
   public func loadTemplate(names: [String], environment: Environment) throws -> Template {
     for path in paths {
       for templateName in names {
-        let templatePath = try path.safeJoin(path: Path(templateName))
+        let templatePath = URL(fileURLWithPath: templateName, relativeTo: URL(fileURLWithPath: path))
+        if !templatePath.withUnsafeFileSystemRepresentation({ String(cString: $0!) }).hasPrefix(path) {
+          throw SuspiciousFileOperation(basePath: path, path: templateName)
+        }
 
-        if templatePath.exists {
-          let content: String = try templatePath.read()
+        if FileManager.default.fileExists(atPath: templatePath.path) {
+          let content = try String(contentsOf: templatePath)
           return environment.templateClass.init(templateString: content, environment: environment, name: templateName)
         }
       }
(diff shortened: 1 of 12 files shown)


Sources/Stencil/Loader.swift
```diff
@@ -51,27 +52,19 @@ public class FileSystemLoader: Loader, CustomStringConvertible {
...
-    throw TemplateDoesNotExist(templateNames: [name], loader: self)
+    return try loadTemplate(names: [name], environment: environment)
   }
 
   public func loadTemplate(names: [String], environment: Environment) throws -> Template {
     for path in paths {
       for templateName in names {
-        let templatePath = try path.safeJoin(path: Path(templateName))
+        let templatePath = URL(fileURLWithPath: templateName, relativeTo: URL(fileURLWithPath: path))
+        if !templatePath.withUnsafeFileSystemRepresentation({ String(cString: $0!) }).hasPrefix(path) {
```
kylef, line 62: Is this equivilent to the old implementation? There doesn't seem to be an explict asolute path, perhaps this is happening for free by this construct?

There also doesn't appear to be any test cases for `SuspiciousFileOperation`.

compnerd: Yes, it should be equivalent to the previous implementation.  The `pathJoin` is concatenating two path components.  The FSR is going to the canonicalised concatenation of the paths.  The current tests do invoke this method IIRC and do pass.

djbe: I'd rather have this in a tiny extension on `URL` or `FileManager`, instead of inside the `loadTemplate` function.

Sources/Stencil/Loader.swift
```diff
@@ -107,23 +100,11 @@ public class DictionaryLoader: Loader {
...
-    }
-
-    return newPath
-  }
-}
-
 class SuspiciousFileOperation: Error {
-  let basePath: Path
-  let path: Path
+  let basePath: String
```
djbe, line 104: Same as above, I'd prefer a more concrete type.

Sources/Stencil/Template.swift
```diff
@@ -41,19 +40,7 @@ open class Template: ExpressibleByStringLiteral {
       throw NSError(domain: NSCocoaErrorDomain, code: NSFileNoSuchFileError, userInfo: nil)
     }
 
-    try self.init(URL: url)
-  }
-
-  /// Create a template with a file found at the given URL
-  @available(*, deprecated, message: "Use Environment/FileSystemLoader instead")
```
djbe, line 48: So if we're dropping these because we no longer have the `Path` type, I'm assuming it'll be at least a "minor" version bump (considering we're still on 0.x.x)

Sources/Stencil/Variable.swift
```diff
@@ -111,12 +111,12 @@ public struct Variable: Equatable, Resolvable {
     } else if let string = context as? String {
       return resolve(bit: bit, collection: string)
     } else if let object = context as? NSObject {  // NSKeyValueCoding
-      #if os(Linux)
-        return nil
-      #else
+      #if _runtime(_ObjC)
```
djbe, line 114: As discussed in #341, I don't there's need for this, and we can just use `canImport` for now.

djbe (changes requested):
I'm not completely against dropping PathKit, it hasn't been updated in quite a while, and pulls in Spectre as a test dependency. Thing to note though: most tools/libraries I know that import Stencil also use PathKit for typed paths.

I noticed there's a PR in PathKit to add Windows support. Does such a PR supersede this one?

Sources/Stencil/Loader.swift
```diff
@@ -34,15 +33,17 @@ extension Loader {
...
 // A class for loading a template from disk
 public class FileSystemLoader: Loader, CustomStringConvertible {
-  public let paths: [Path]
+  public let paths: [String]
 
-  public init(paths: [Path]) {
-    self.paths = paths
+  public init(paths: [URL]) {
+    self.paths = paths.map {
+      $0.withUnsafeFileSystemRepresentation { String(cString: $0!) }
```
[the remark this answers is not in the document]
compnerd: No, we cannot.  The `.path` does not give you a *file system representation*.  It gives you a URL path representation.  This just happens to work on some platforms, but is not guaranteed, and does in fact break on Windows.  The filesystem operations operate on the file system representation, which is only provided as a C-string.

Sources/Stencil/Template.swift
compnerd: Yeah, that seems fair.

Sources/Stencil/Variable.swift
compnerd: Yeah, that commit needs to be dropped, I couldn't easily do so though.

compnerd:
> I'm not completely against dropping PathKit, it hasn't been updated in quite a while, and pulls in Spectre as a test dependency. Thing to note though: most tools/libraries I know that import Stencil also use PathKit for typed paths.
> 
> I noticed there's a [PR in PathKit to add Windows support](https://github.com/kylef/PathKit/pull/86). Does such a PR supersede this one?

No, it does not.  That change does allow compilation, but pretty quickly falls apart.  Fixing that would require a pretty invasive rewrite that I don't think is valuable (at least not sufficiently that I would be willing to do that work).

Sources/Stencil/Loader.swift
```diff
@@ -34,15 +33,17 @@ extension Loader {
 
 // A class for loading a template from disk
 public class FileSystemLoader: Loader, CustomStringConvertible {
-  public let paths: [Path]
+  public let paths: [String]
```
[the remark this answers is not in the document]
compnerd: Sure, I can change this to `[URL]`.

kylef:
> Porting PathKit turned out to be more difficult than anticipated and it was easier to remove that in favour of Foundation.

Wondering what makes it so difficult, is it a subset of APIs are hard to port like glob or something more fundamental?

PR #345, opened 2023-11-29
Caching loaded template lines to improve performance

pavel-trafimuk:
Currently, Stencil load and parse template file for each new output file generation,
this improve help to reduce repetitive logic and significantly improve performance (on 30-50%)

### DIFF
--- a/Sources/Stencil/Context.swift
+++ b/Sources/Stencil/Context.swift
@@ -9,7 +9,7 @@ public class Context {
   var dictionaries: [[String: Any?]]
 
   /// The context's environment, such as registered extensions, classes, …
-  public let environment: Environment
+  public var environment: Environment
 
   init(dictionaries: [[String: Any?]], environment: Environment) {
     self.dictionaries = dictionaries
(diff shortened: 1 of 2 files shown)


art-divin:
cc @djbe

tothambrus11:
Hello, can somebody review this PR? This change seems to be very useful for our project at Hylo, where we generate thousands of pages for the documentation of the standard library. 

To my understanding, this would be a breaking change, as the caches require the generator functions to be mutating. In swift there might be a way to opt out of the strict value semantics and make the cache a mutable field that is not part of the object's value, something like the `mutable` keyword for struct members in C++ that allows the field to be mutated from a const function, thus allowing the introduction of caches with const API.

pavel-trafimuk:
@tothambrus11 seems like repo is not maintenaning anymore. this improved already implemented in https://github.com/art-divin/StencilSwiftKit.git and https://github.com/art-divin/StencilSwiftKit.git and finally in Sourcery and it was checked by time.

